PR #39, opened 2019-02-28
Add missing properties to case and workflow factory, add test

czosel:
... and also update the schema.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,6 +92,7 @@
     "eslint-plugin-ember": "6.1.0",
     "eslint-plugin-node": "8.0.1",
     "eslint-plugin-prettier": "2.6.0",
+    "graphql-cli": "3.0.11",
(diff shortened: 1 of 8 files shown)


czosel:
@kaldras ready for review

package.json
```diff
@@ -92,6 +92,7 @@
     "eslint-plugin-ember": "6.1.0",
     "eslint-plugin-node": "8.0.1",
     "eslint-plugin-prettier": "2.6.0",
+    "graphql-cli": "3.0.11",
```
kaldras: Where do you use this? Or why is this needed?

czosel: This is needed to run `graphql schema-update` - i forgot to migrate this dependency from ember-caluma-utils.

kaldras: Ok, thanks for the explanation :+1:

PR #42, opened 2019-03-05
Add support for adding TableQuestions in form builder

czosel:


czosel:
@fkm Ready for review.

addon/templates/components/cfb-form-editor/question.hbs
```diff
@@ -3,7 +3,7 @@
     {{uk-spinner ratio=2}}
   </div>
 {{else}}
-  {{#with data.lastSuccessful.value.firstObject.node as |model|}}
+  {{#with model as |model|}}
```
fkm: Habe erst mal das Diff überflogen. Braucht es das `with` hier überhaupt noch?

Ist mehr eine Verständnisfrage.

czosel: English, please :wink: 

I kept with because there's an "else" branch (if model is null) all the way below. Probably better to use `if/else` explicitly though.

fkm: Noted :-)

Ok. Sounds reasonable!

czosel: I'll make the change in #44, since it's based on this PR.

PR #43, opened 2019-03-05
Custom widgets for single and multiple choice questions

fkm:
Powered by [Ember Power Select](https://ember-power-select.com/)

* Current threshold for search field is at 3 choices.
* Currently not styled through UIkit.
* `searchMessage` seems to be overridden by `noMatchesMessage` (see [API reference](https://ember-power-select.com/docs/api-reference/))

### DIFF
--- a/addon/components/cf-field/input/powerselect.js
+++ b/addon/components/cf-field/input/powerselect.js
@@ -22,7 +22,7 @@ export default Component.extend({
     "multiple",
     "field.question.{choiceOptions,multipleChoiceOptions}.edges",
     function() {
-      let options = this.get("multiple")
+      const options = this.get("multiple")
         ? this.get("field.question.multipleChoiceOptions")
         : this.get("field.question.choiceOptions");
 
@@ -67,8 +67,8 @@ export default Component.extend({
   }),
 
   placeholder: computed("multiple", function() {
-    let suffix = this.get("multiple") ? "multiple" : "single";
-    let path = `caluma.form.power-select.placeholder-${suffix}`;
+    const suffix = this.get("multiple") ? "multiple" : "single";
+    const path = `caluma.form.power-select.placeholder-${suffix}`;
     return this.get("intl").t(path);
   }),
 
--- a/addon/components/cf-field/input/powerselect.js
+++ b/addon/components/cf-field/input/powerselect.js
@@ -13,11 +13,12 @@ export default Component.extend({
   }),
 
   choices: computed(
+    "multiple",
     "field.question.{choiceOptions,multipleChoiceOptions}.edges",
     function() {
-      let options =
-        this.get("field.question.choiceOptions") ||
-        this.get("field.question.multipleChoiceOptions");
+      let options = this.get("multiple")
+        ? this.get("field.question.multipleChoiceOptions")
+        : this.get("field.question.choiceOptions");
 
       return options.edges.map(edge => edge.node);
     }
--- a/addon/components/cf-field/input/powerselect.js
+++ b/addon/components/cf-field/input/powerselect.js
@@ -35,8 +35,20 @@ export default Component.extend({
     function() {
       let key = this.get("field.answer._valueKey");
       let path = `field.answer.${key}`;
+      let answers = this.get(path);
+      let selection = null;
 
-      return this.get(path);
+      if (answers !== null) {
+        selection = this.get("choices").filter(choice => {
+          return answers.includes(choice.slug);
+        });
+
+        if (key === "stringValue") {
+          selection = selection[0];
+        }
+      }
+
+      return selection;
     }
   ),
 
@@ -63,9 +75,15 @@ export default Component.extend({
   actions: {
     change: function(choices) {
       let key = this.get("field.answer._valueKey");
-      let path = `field.answer.${key}`;
+      let value = null;
 
-      this.set(path, choices);
+      if (key === "listValue") {
+        value = choices.map(choice => choice.slug);
+      } else if (choices !== null) {
+        value = choices.slug;
+      }
+
+      this.onSave(value);
     }
   }
 });
--- a/addon/components/cf-field/input/powerselect.js
+++ b/addon/components/cf-field/input/powerselect.js
@@ -71,6 +71,9 @@ export default Component.extend({
       } else if (choices !== null) {
         value = choices.slug;
       }
+      // ELSE will never be taken as long as we don't allow for empty
+      // selections in single choice fields. Empty selections must first be
+      // implemented/allowed by the backend.
 
       this.onSave(value);
     }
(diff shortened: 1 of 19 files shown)


addon/templates/components/cfb-form-editor/question.hbs
```diff
@@ -121,6 +121,18 @@
...
+        {{f.input
+          name="widgetType"
+          label=(t "caluma.form-builder.question.widgetType")
+          required=true
+          class="uk-flex uk-flex-between uk-flex-column"
+
+          type="select"
+          optionTargetPath="value"
+          optionLabelPath="label"
+          options=(array (hash value="default" label="Default") (hash value="powerselect" label="Dropdown"))
```
czosel: If you use `radio` instead of `default` you can drop the conditional in cf-field/input.js:
`(customtype && customtype !== "default" && customtype)`

czosel: I think calling it "Radio" instead of "Default" also makes more sense for the user, especially if the question type is "Choice" and not "Radio".

addon/components/cf-field/input/powerselect.js
```diff
@@ -0,0 +1,89 @@
...
+        selection = this.get("choices").filter(choice => {
+          return answers.includes(choice.slug);
+        });
+
+        if (key === "stringValue") {
+          selection = selection[0];
+        }
+      }
+
+      return selection;
```
czosel: I would have written this like that:
```js
      const key = this.get("field.answer._valueKey");
      const answers = this.get(`field.answer.${key}`);

      if (!answers) {
        return null;
      }
      const selection = this.get("choices").filter(choice => {
        return answers.includes(choice.slug);
      });

      return key === "stringValue" ? selection[0] : selection:
```

* prefer-const-over-let (we should add this to our eslint config)
* nest code as little as possible / return early
* as little assignment as possible

addon/components/cf-field/input/powerselect.js
```diff
@@ -0,0 +1,89 @@
...
+  ),
+
+  componentName: computed("multiple", function() {
+    let name = "power-select";
+
+    if (this.get("multiple")) {
+      name += "-multiple";
+    }
+
+    return name;
```
czosel:
```js
return this.get("multiple") ? "power-select-multiple" : "power-select"
```
looks more readable and concise to me, even though the strings have a large part in common.

addon/components/cf-field/input/powerselect.js
```diff
@@ -0,0 +1,89 @@
...
+  }),
+
+  searchEnabled: computed("choices", function() {
+    return this.get("choices").length > 10;
+  }),
+
+  placeholder: computed("multiple", function() {
+    let suffix = this.get("multiple") ? "multiple" : "single";
+    let path = `caluma.form.power-select.placeholder-${suffix}`;
+    return this.get("intl").t(path);
```
czosel: * use const instead of let
* probably better to drop `path`

addon/components/cf-field/input/powerselect.js
```diff
@@ -0,0 +1,89 @@
...
+  }),
+
+  actions: {
+    change: function(choices) {
+      let key = this.get("field.answer._valueKey");
+      let value = null;
+
+      if (key === "listValue") {
```
czosel: Maybe rather to check if `choices` is an array?

tests/integration/components/cf-field/input/powerselect-test.js
```diff
@@ -0,0 +1,136 @@
...
+        __typename: "ChoiceQuestion",
+        choiceOptions: {
+          edges: [
+            { node: { slug: "option-1", label: "Option 1" } },
+            { node: { slug: "option-2", label: "Option 2" } },
+            { node: { slug: "option-3", label: "Option 3" } }
+          ]
+        }
+      }
+    });
```
czosel: maybe better to set `singleChoiceField` and `multipleChoiceField` in beforeEach to avoid repeating the lengthy `set` call in each test?

translations/de-de.yaml
```diff
@@ -5,6 +5,15 @@ dummy:
...
   form:
     optional: "Optional"
+
+    power-select:
+      placeholder-single: "Wählen Sie eine Option aus"
+      placeholder-multiple: "Wählen Sie eine oder mehrere Optionen aus"
+      options-loading: "Lade Optionen..."
+      options-empty: "Keine Optionen vorhanden"
+      search-placeholder: "Hier tippen um zu suchen"
+      search-empty: "Keine Optionen entsprechen der Suche"
```
czosel:
```suggestion
      search-empty: "Keine Optionen gefunden"
```

czosel:
Mostly nitpicking, looks good in general! :+1:

addon/templates/components/cfb-form-editor/question.hbs
fkm: The problem with a specific default type is that the widget is also used for the MultipleChoiceQuestion type.

addon/components/cf-field/input/powerselect.js
fkm: I personally prefer to have only one return per function where possible. But I can live with multiple returns :-)

What options would you recommend for prefer-const?
https://eslint.org/docs/rules/prefer-const

addon/components/cf-field/input/powerselect.js
fkm: Makes sense.

addon/components/cf-field/input/powerselect.js
fkm: Dropping path would lead to a rather long line :-\

addon/components/cf-field/input/powerselect.js
fkm: I had that at first and then decided that using the key is 'better'. But revisiting it, I agree that checking choices is probably the better way.

tests/integration/components/cf-field/input/powerselect-test.js
fkm: Good idea!

addon/templates/components/cfb-form-editor/question.hbs
czosel: You could make the options dynamic though, right?

fkm: In theory definitely. But I'm not sure it will make the code shorter/better/easier. Although having the hash defined in the component instead of the template would be cleaner. I'm just not sure about the dynamic content.

addon/components/cf-field/input/powerselect.js
czosel: Right - in my opinion the mental overhead of tracking the value of a variable is higher than of finding all the `return` statements in a function. Also we try to write our JS in a "functional" style which means avoiding mutable variables.

See also: https://stackoverflow.com/questions/36707/should-a-function-have-only-one-return-statement

For prefer-const I'd just stick with the default options, resp. not set any specific options.

addon/components/cf-field/input/powerselect.js
czosel: Ok, let's leave it then.

czosel:
Looks all good now! Only codecov still seems to be unhappy ...

PR #46, opened 2019-03-08
FIX missing dependency transmission

fkm:
**Do not merge this before the consuming apps have been changed!**

This fixes the missing translations by requesting the `intl` service from the host app.

Additionally, this change adds the dependency to the dummy app to avoid having an error, when running the add-on standalone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,24 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+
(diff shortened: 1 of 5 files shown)


czosel:
I'd say we should add:

- [ ] A CHANGELOG file (see [this example](https://github.com/adfinis-sygroup/ember-validated-form/blob/master/CHANGELOG.md))
- [ ] Updated [the docs](https://caluma.io/ember-caluma/docs) describing the extra dependency

PR #50, opened 2019-03-11
Change question list to show archived

fkm:
This only applies to the list of questions already assigned to a form.

We still don't show - and don't intend to - archived questions when adding new questions.

### DIFF
--- a/addon/gql/queries/search-form-question.graphql
+++ b/addon/gql/queries/search-form-question.graphql
@@ -1,11 +1,11 @@
 #import "ember-caluma/gql/fragments/question-info"
 
-query SearchFormQuestion($slug: String!, $search: String) {
+query SearchFormQuestion($slug: String!, $search: String, $archived: Boolean) {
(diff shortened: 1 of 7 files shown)


czosel:
Looks good! Just tests are still failing.

addon/gql/queries/search-form-question.graphql
```diff
@@ -1,11 +1,11 @@
 #import "ember-caluma/gql/fragments/question-info"
 
-query SearchFormQuestion($slug: String!, $search: String) {
+query SearchFormQuestion($slug: String!, $search: String, $archived: Boolean) {
   allForms(slug: $slug) {
     edges {
       node {
         slug
-        questions(search: $search, isArchived: false) {
+        questions(search: $search, isArchived: $archived) {
```
fkm: This here seems to be the culprit. If I change this back to a static `false` the refresh of the question list works as expected.

czosel:
@fkm Tests are still failing?

PR #63, opened 2019-03-15
Add hidden field in "advanced settings" block

### LINKED ISSUE #48: Form builder: Show "isHidden" field
For developing a complex form configuration, `isHidden`expressions currently have to be set using the API. This is tedious, because the API requires several fields which are necessary for the initial create, but not actually needed for the update (e.g. options in ChoiceQuestions).

Maybe we could add a text field for `isHidden` which is part of a "advanced settings" section of the question form? The advanced settings section could be hidden by default and only shown when activated (toggled), maybe also featuring a little warning like "Use at your own risk..."

czosel:
Fixes #48

### DIFF
--- a/addon/templates/components/cfb-form-editor/question.hbs
+++ b/addon/templates/components/cfb-form-editor/question.hbs
@@ -154,6 +154,15 @@
         class="uk-flex uk-flex-between uk-flex-column"
       }}
 
+      <details>
+        <summary>{{t "caluma.form-builder.question.advancedSettings"}}</summary>
+        {{f.input
+          type="text"
+          label=(t "caluma.form-builder.question.isHidden")
+          name="isHidden"
+        }}
+      </details>
+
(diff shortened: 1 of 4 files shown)


addon/templates/components/cfb-form-editor/question.hbs
```diff
@@ -154,6 +154,15 @@
         class="uk-flex uk-flex-between uk-flex-column"
       }}
 
+      <details>
```
fkm: Not sure if you checked the [browser support](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#Browser_compatibility).

czosel: Oh damn, very good point. I'll look for another solution.

czosel: Fixed

PR #66, opened 2019-03-19
ADD component overrides and crossreferences

fkm:
* Specify overriding components via slug.
* Reference other questions via slug.

### DIFF
--- a/addon/components/cf-form.js
+++ b/addon/components/cf-form.js
@@ -55,5 +55,22 @@ export default Component.extend(ComponentQueryManager, {
       (this.get("data.lastSuccessful.value") &&
         this.documentStore.find(this.get("data.lastSuccessful.value")))
     );
-  }).readOnly()
+  }).readOnly(),
+
+  actions: {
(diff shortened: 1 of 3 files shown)


czosel:
Can you update the documentation for this?

Also, I'm not really happy with `crossReference` - it looks more like a custom functionality instead of something that should go in the addon. Can't you call the fields `save` method without the extra action, just based on the `field` property which is passed into the custom component?

fkm:
I found a way.

czosel:
So you'll open a new PR for the component override part?

fkm:
I've assigned you a pull request in the project repository. The cross-reference part is now done through field.document.fields and doesn't require any change to ember-caluma.

czosel:
Awesome! But you'll need a change here for the override part, right?

fkm:
You're right! My bad!

czosel:
Can you add a short test and some docs for this?

czosel:
Merging for our deployment, created #73 to keep track of the missing parts.

PR #68, opened 2019-03-21
Add Pikaday datepicker

### LINKED ISSUE #67: Add date question
Backend support for this landed in https://github.com/projectcaluma/caluma/pull/333

fkm:
There remains an error, when selecting a date.

I don't even know where the error comes from...

> "Assertion Failed: Cannot call get with 'errorFormat' on an undefined object."

I will provide a fix and remove the "WIP" as soon as possible.

Resolves #67


fkm:
@czosel Hatte gestern nochmals ein Runde debugged und hab die Tets noch nicht durch.

Aber funktionalität ist soweit fertig.

czosel:
conflicts ... :unamused:

ember-cli-build.js
```diff
@@ -15,6 +15,13 @@ module.exports = function(defaults) {
     }
   });
 
+  app.import("node_modules/moment/locale/de.js");
+  app.import("node_modules/moment/locale/de-ch.js");
+  app.import("node_modules/moment/locale/it.js");
+  app.import("node_modules/moment/locale/it-ch.js");
+  app.import("node_modules/moment/locale/fr.js");
+  app.import("node_modules/moment/locale/fr-ch.js");
```
czosel: Does this affect consuming apps or just the dummy app?

tests/integration/components/cf-form-test.js
```diff
@@ -140,6 +146,11 @@ module("Integration | Component | cf-form", function(hooks) {
       slug: "file-question",
       type: "FILE"
     });
+    this.server.create("question", {
+      formIds: [form.id],
+      slug: "date-question",
+      type: "DATE"
+    });
```
czosel: I suppose this is commented out because it's causing issues? Maybe add a comment explaining the place where you got stuck?

package.json
```diff
@@ -90,7 +91,8 @@
     "ember-export-application-global": "2.0.0",
     "ember-load-initializers": "2.0.0",
     "ember-maybe-import-regenerator": "0.1.6",
-    "ember-power-select": "2.3.1",
+    "ember-power-select": "2.2.3",
```
czosel: Was the downgrade of ember-power-select intentional?

ember-cli-build.js
fkm: If I understood Ember CLI correctly, this is for all apps. But I'd have to test it to be sure.

tests/integration/components/cf-form-test.js
fkm: Yes. That's the 'skip'. I'll add a comment.

package.json
fkm: Good catch!

Issue #60, opened 2019-03-14
Form builder: Add namespace for slugs


kaldras:
While initializing the form builder it should be possible to define a namespace which is then added to each slug that is created in this session. For example:

`namespace: "foobar"`

Question slug of question "blubber" should be: `foobar-blubber`

It would be nice to see the namespace as non-editable text before the slug field, so the user knows how the final slug looks.

fkm:
I have yet to find a way to pass the user/group (i.e. session) to the addon.

I could use a dependency injection to pass the OpenID session to the addon and the application configuration to define what part of the session should be used (eg. via Jexl string). The problem with the dependency is that it will fall flat when the host application doesn't provide a session for injection.

fkm:
Maybe via a [custom authenticator](https://github.com/simplabs/ember-simple-auth/tree/master#implementing-a-custom-authenticator). I will have a look.

fkm:
After looking at everything I could think of, I currently only see the solution in attaching the namespace to `window`. But I'm open for suggestions.

PR #97, opened 2019-03-28
Add option service

fkm:
To share data between host and addon at runtime.

E.g. namespace for question slugs.

### DIFF
--- a/addon/components/cfb-form-editor/question.js
+++ b/addon/components/cfb-form-editor/question.js
@@ -42,7 +42,7 @@ export default Component.extend(ComponentQueryManager, {
 
   notification: service(),
   intl: service(),
-  options: service(),
+  calumaOptions: service(),
 
   possibleTypes: computed(function() {
     return Object.keys(TYPES).map(value => ({
@@ -164,8 +164,8 @@ export default Component.extend(ComponentQueryManager, {
     return model;
   }),
 
-  namespace: computed("options.namespace", function() {
-    return slugify(this.options.get("namespace") || "") || null;
+  namespace: computed("calumaOptions.namespace", function() {
+    return slugify(this.calumaOptions.get("namespace") || "") || null;
   }),
 
   _getIntegerQuestionInput(changeset) {
--- a/addon/components/cfb-form-editor/question.js
+++ b/addon/components/cfb-form-editor/question.js
@@ -53,17 +53,19 @@ export default Component.extend(ComponentQueryManager, {
 
   addSlug() {
     const namespace = this.options.get("namespace");
-
     const input = this.element.querySelector('[name="slug"]');
-    const span = document.createElement("span");
-    const parent = input.parentElement;
 
-    Object.assign(span, {
-      className: "slugnamespace-slug",
-      innerHTML: namespace === undefined ? "" : `${namespace}-`
-    });
-    parent.classList.add("slugnamespace");
-    parent.insertBefore(span, input);
+    if (namespace && input) {
+      const span = document.createElement("span");
+      const parent = input.parentElement;
+
+      Object.assign(span, {
+        className: "slugnamespace-slug",
+        innerHTML: `${namespace}-`
+      });
+      parent.classList.add("slugnamespace");
+      parent.insertBefore(span, input);
+    }
   },
 
   async didReceiveAttrs() {
@@ -71,8 +73,6 @@ export default Component.extend(ComponentQueryManager, {
 
     await this.get("data").perform();
     await this.get("availableForms").perform();
-
-    this.options.set("namespace", "foo");
 
     if (!this.get("slug")) {
       this.addSlug();
--- a/addon/components/cfb-form-editor/question.js
+++ b/addon/components/cfb-form-editor/question.js
@@ -43,16 +43,13 @@ export default Component.extend(ComponentQueryManager, {
 
   notification: service(),
   intl: service(),
+  options: service(),
 
   possibleTypes: computed(function() {
     return Object.keys(TYPES).map(value => ({
       value,
       label: this.intl.t(`caluma.form-builder.question.types.${value}`)
     }));
-  }),
-
-  slugPrefix: computed("form", function() {
-    return this.get("form") + "-";
   }),
 
   prepareSlug() {
@@ -135,11 +132,13 @@ export default Component.extend(ComponentQueryManager, {
 
   data: task(function*() {
     if (!this.get("slug")) {
+      const slug = this.options.get("namespace");
+
       return A([
         {
           node: {
             label: "",
-            slug: this.slugPrefix,
+            slug: slug === undefined ? "" : `${slug}-`,
             description: "",
             isRequired: "false",
             isHidden: "false",
@@ -345,7 +344,10 @@ export default Component.extend(ComponentQueryManager, {
       changeset.set("label", value);
 
       if (!this.get("slug")) {
-        const slug = this.slugPrefix + slugify(value);
+        let prefix = this.options.get("namespace");
+        prefix = prefix === undefined ? "" : `${prefix}-`;
+
+        const slug = prefix + slugify(value);
         changeset.set("slug", slug);
         this.get("validateSlug").perform(slug, changeset);
       }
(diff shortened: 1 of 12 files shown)


fkm:
Resolves #60

addon/components/cfb-form-editor/question.js
```diff
@@ -50,11 +52,69 @@ export default Component.extend(ComponentQueryManager, {
...
+    });
+
+    $input.off("focus.slug click.slug select.slug");
+    $input.on("focus.slug click.slug select.slug", event => {
+      const input = event.target;
+      if (input.selectionStart <= length) {
+        input.selectionStart = length;
+      }
+    });
+  },
```
czosel: What is this doing? :sweat_smile:

fkm: Keeping the user from modifying the namespace while still allowing him to edit the rest as expected.

fkm: I thought about overlaying another element. But that is not as straight forward as I initially thought as the namespace has a variable width which would have to be somehow taken into account. The easiest solution in the end was to have the whole string (namespace and slug) in the same input and restrict modification in the UI.

There's a small bug when selecting text from the first editable character onward. But with normal usage this should not be an immediate issue.

czosel: Hmm, why not just the low-tech approach?
```html
<div>
  {{namespace}}
  <input type="text" />
</div>
```

fkm: Hmmm.. Flexbox might make this possible. I'll look into that. With traditional methods there would be no way to size the input element to fill the remaining space after the namespace string.

czosel: Sure? I'd expect this to work even without flexbox with just a `float`. Anyway, flexbox or the UIkit grid are definitely better.

fkm: The problem with floating is that you usually get into sizing complications.

But I just stumbled onto another issue. The input field is created by ember-validated-form. Not sure if it is possible to amend the generated markup. But I'll look into that.

czosel: The way the required field is positioned might offer some guidance:
```html
      <div uk-grid class="uk-grid-small uk-margin">
        <div class="uk-width-expand">
          {{f.input
            type="text"
            label=(t "caluma.form-builder.question.slug")
            name="slug"
            required=true
            disabled=slug
            on-update=(action "updateSlug")
          }}
        </div>

        <div class="uk-width-auto uk-flex">
          {{f.input
            name="isRequired"
            label=(t "caluma.form-builder.question.isRequired")
            required=true
            renderComponent=(component "cfb-jexl-boolean-toggle-switch" size="small")
            class="uk-flex uk-flex-between uk-flex-column"
          }}
        </div>
      </div>
```

fkm: Something like this?

namespace-slug

addon/components/cfb-form-editor/question.js
```diff
@@ -50,11 +51,44 @@ export default Component.extend(ComponentQueryManager, {
...
+
+    if (!this.get("slug")) {
+      this.addSlug();
+    }
+  },
+
+  willDestroyElement() {
+    this.$('[name="slug"]').off(
+      "keypress.slug keydown.slug focus.slug click.slug select.slug"
+    );
```
czosel: This is not needed anymore, is it?

addon/components/cfb-form-editor/question.js
```diff
@@ -50,11 +51,44 @@ export default Component.extend(ComponentQueryManager, {
...
+    }
+  },
+
   async didReceiveAttrs() {
     this._super(...arguments);
 
     await this.get("data").perform();
     await this.get("availableForms").perform();
+
+    if (!this.get("slug")) {
```
czosel: Why are we only adding the prefix if there is no `slug`?

czosel: Ah nevermind, found the `this.set("slug", ...)` call below

addon/components/cfb-form-editor/question.js
```diff
@@ -196,6 +230,14 @@ export default Component.extend(ComponentQueryManager, {
 
   submit: task(function*(changeset) {
     try {
+      if (!this.get("slug")) {
+        const namespace = this.options.get("namespace");
+        if (namespace !== undefined) {
```
czosel: `if (namespace)` should be enough

addon/components/cfb-form-editor/question.js
fkm: It is not. Forgot to remove that.

addon/components/cfb-form-editor/question.js
```diff
@@ -42,6 +42,7 @@ export default Component.extend(ComponentQueryManager, {
 
   notification: service(),
   intl: service(),
+  options: service(),
```
czosel: Maybe you could simplify your code by adding
```
namespace: reads("options.namespace")
```
(didn't test if that works with properties on services, but i guess it should)

addon/components/cfb-form-editor/question.js
fkm: It should. Except when it's "0" or something else that happens to be falsy when converted. As long as I don't know what `namespace` could be, I'd rather not use a truthy condition.

czosel:
Might be good idea to add an example implementation that uses the options service to the dummy app?

addon/components/cfb-form-editor/question.js
fkm: Interesting! I'll try that :-)

addon/components/cfb-form-editor/question.js
czosel: I'd rather impose some conditions on the values that can be used as namespace. It should always be a string for one, which solves the issue with "0". It can't be an empty string, otherwise it'll be treated as if no namespace is present (which makes total sense). Also, are you currently making sure that the concatenated result (namespace + actual slug) is slugified together?

fkm: Namespace is now a computed property and an empty string if not present.

addon/components/cfb-form-editor/question.js
```diff
@@ -56,16 +56,19 @@ export default Component.extend(ComponentQueryManager, {
    * This uses manual DOM manipulation to avoid adding a single-use component.
    */
   addSlug() {
-    const namespace = this.options.get("namespace");
     const input = this.element.querySelector('[name="slug"]');
 
-    if (namespace && input) {
+    if (
+      this.namespace.length &&
```
czosel: Why `.length`?

addon/components/cfb-form-editor/question.js
```diff
@@ -135,6 +164,10 @@ export default Component.extend(ComponentQueryManager, {
     return model;
   }),
 
+  namespace: computed("options.namespace", function() {
+    return slugify(this.options.get("namespace") || "");
```
czosel: You could also do something like 
```suggestion
    return slugify(this.options.get("namespace")) || null;
```
Or is `slugify` throwing if undefined is passed? If yes, then you'd have to keep `|| ""` but could still add the `|| null`.

fkm: Yeah. Slugify needs a string. Otherwise, it throws at `.toLowerCase()`.
https://github.com/projectcaluma/ember-caluma/blob/master/addon/utils/slugify.js#L4

czosel: Oh wow, I'm quite surprised that we implimented this ourselves in the first place. Anyway, in that case it would be 
```suggestion
    return slugify(this.options.get("namespace") || "") || null;
```

addon/components/cfb-form-editor/question.js
fkm: To check if the namespace is not empty. I usually write `.length > 0` but tried to be as concise as possible.

addon/components/cfb-form-editor/question.js
czosel: Or you just patch slugify to return `undefined` if no argument is given :wink:

fkm: Why `null`?

addon/components/cfb-form-editor/question.js
czosel: Since empty strings are falsy, you can drop `.length` alltogether.

czosel: And with the patch discussed below `namespace` would even be `null`.

czosel:
Hmm, tests still seem to be failing ...

PR #103, opened 2019-04-01
Subform navigation

czosel:
Depends on https://github.com/projectcaluma/caluma/pull/349

### DIFF
--- a/addon/components/cf-field.js
+++ b/addon/components/cf-field.js
@@ -44,14 +44,12 @@ export default Component.extend({
 
     yield this.field.validate.perform();
 
-    if (this.field.isInvalid) {
-      return;
+    try {
+      const response = yield this.field.save.unlinked().perform();
+      answer.setProperties(response);
+      return response;
+    } catch (e) {
+      // that's ok
     }
-
-    const response = yield this.field.save.unlinked().perform();
-
-    answer.setProperties(response);
-
-    return response;
   }).restartable()
 });
--- a/addon/components/cf-field/errors.js
+++ b/addon/components/cf-field/errors.js
@@ -4,5 +4,5 @@ import layout from "../../templates/components/cf-field/errors";
 export default Component.extend({
   layout,
   tagName: "span",
-  classNames: ["uk-text-small", "uk-text-danger"]
+  classNames: ["uk-text-small", "uk-text-danger", "validation-errors"]
 });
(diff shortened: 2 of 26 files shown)


fkm:
I get the following error:
> Cannot read property 'fields' of undefined
    at cf-navigation.js:37

The property `field.childDocument` is set but to `undefined`.

czosel:
@fkm would be awesome if you could take another look :slightly_smiling_face:

fkm:
I still get the same error. Here's what I've got in my backend:

```
{
  "data": {
    "allForms": {
      "edges": [
        {
          "node": {
            "slug": "form-2",
            "questions": {
              "edges": [
                {
                  "node": {
                    "slug": "q1",
                    "__typename": "TextQuestion",
                    "forms": {
                      "edges": [
                        {
                          "node": {
                            "slug": "form-2"
                          }
                        }
                      ]
                    }
                  }
                }
              ]
            }
          }
        },
        {
          "node": {
            "slug": "form-1",
            "questions": {
              "edges": [
                {
                  "node": {
                    "slug": "f1",
                    "__typename": "FormQuestion",
                    "forms": {
                      "edges": [
                        {
                          "node": {
                            "slug": "form-1"
                          }
                        }
                      ]
                    }
                  }
                },
                {
                  "node": {
                    "slug": "q2",
                    "__typename": "TextQuestion",
                    "forms": {
                      "edges": [
                        {
                          "node": {
                            "slug": "form-1"
                          }
                        }
                      ]
                    }
                  }
                }
              ]
            }
          }
        }
      ]
    }
  }
}
```

Could the problem arise from having the sub form attached to every question? (i.e. q1 and q2)

I managed to have the sub form work at some point. But then I added more stuff and everything broke. I try to find out, which case works.

czosel:
Yeah, having the same subform attached to multiple FormQuestions could definitely be the issue. Can you test with different subforms?

fkm:
Attaching the sub forms to question types other than FormQuestion is not something I did through the UI. Seems to be something GraphQL does in the background. Either on save or fetch.

Here's the query I used to get the previous result:

```
query {
  allForms {
    edges {
      node {
        slug
        questions {
          edges {
            node {
              slug
              forms {
                edges {
                  node {
                    slug
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

fkm:
But at the moment it doesn't display anything. The document/form is returned by the backend but the form is missing from the frontend. I'll have a look at the code.

PR #104, opened 2019-04-01
Add widget overrides from dropdown

fkm:
Depends on #97 

I included the Power Select override as an example on how to use this change to handle widget overrides that are included in the framework. With that we could potentially remove the special dropdown on choice and multiplechoice questions.

The format for the overrides is as follows:
```js
{
  label: "Will be shown in the Dropdown",
  component: "the-component-path",
  types: ["optional", "type", "restriction"]
}
```

Which could look like this:
```js
{
  label: this.intl.t("caluma.form-builder.question.widgetOverrides.powerselect"),
  component: "cf-field/input/powerselect",
  types: ["choice", "multiplechoice"]
}
```

I thought about renaming `widgetOverrides` to just `overrides` but am afraid that would be too generic to avoid a naming collision in the future. I'm all ears for another name - preferably one work.

I shortened and lower-cased the type names. Not sure if that was a good idea. The reasoning was to keep the configuration as short as possible. But with the path string for `this.intl.t` that became sort of irrelevant.


fkm:
Ah. I almost forgot! Why this is a "WIP":

I haven't found a way to have the dropdown watch changes from the changeset. I'm sure there's a way but I wanted to give you something early because of the potential replacement for the "custom widget" dropdown.

Anyways. You might have to refresh the page to see the correct overrides in the dropdown. A fix will follow.

addon/services/options.js
```diff
@@ -49,5 +49,53 @@ export default Service.extend({
...
+   *
+   * @param {Object} override The additional override.
+   * @param {String} override.label The text displayed in the form-builder dropdown.
+   * @param {String} override.component The path/name of the overriding component.
+   * @param {Array} [override.types] An optional question type restriction.
+   * @return {Void}
+   */
+  registerComponentOverride(override) {
+    const index = this._overrides.findIndex(
+      datum => datum.component === override.component
```
czosel: datum?

addon/services/options.js
```diff
@@ -13,17 +13,17 @@ export default Service.extend({
 
   init() {
     this._super(...arguments);
-    this._data = new EmberObject();
 
-    this._data.set("overrides", [
+    this._data = new EmberObject();
+    this._overrides = [
```
czosel: Wouldn't it be much simpler to use an object indexed by component name instead of a list? Then you can replace all the finding and splicing by simple assignments and member checks.

addon/services/options.js
```diff
@@ -49,5 +49,53 @@ export default Service.extend({
...
+    }
+  },
+
+  /**
+   * Returns the registered overrides, optionally filtered by a predicate.
+   *
+   * @param {Function} [predicate] The predicate function used by the filter.
+   * @return {Array} The list of overrides matching the predicate.
+   */
+  getComponentOverrides(predicate = () => true) {
```
czosel: Does this API provide any real benefit over 
`options.getComponentOverrides().filter(...)`?

addon/services/options.js
fkm: Oops. Forgot to update that one after switching from `_data` to `_overrides`. Should be `override`.

PS: datum (sing) -> data (pl)

addon/templates/components/cfb-form-editor/question.hbs
```diff
@@ -3,12 +3,9 @@
     {{uk-spinner ratio=2}}
   </div>
 {{else}}
-  {{#with model as |model|}}
+  {{#with model}}
```
czosel: This looked strange before, and now looks even stranger :smile: Shall we replace it with a simple `{{if}}`?

fkm:
Thing is rebased to upstream/master.

The only thing left is to decide whether or not to ditch the `widgetType` (for Choice and MultipleChoice) and what to rename `widgetOverride` to - if at all.

kaldras:
An update of the documentation with the `caluma-options` service would be nice (I forgot to mention this in the last PR..) otherwise the user is wondering why the engine not works.

fkm:
Added the `caluma-options` service to the dependency snippet and amended the existing YUIDoc comments.

I still have no idea, why the documentation for the following methods is not included in the documentation.
* `registerComponentOverride`
* `unregisterComponentOverride`
* `getComponentOverrides`

czosel:
Same here, can you merge in master?

czosel:
All in all, this looks really good already. After trying it out, I do think that it would be best to merge "Widget type" and "Custom Widget" into one field. The power-select rendering would then just be a pre-set override. Or does anything speak against that @fkm ?

fkm:
I don't see a problem with merging the two features.

PR #106, opened 2019-04-02
Document store: use map instead of array, add save method

czosel:
Before, it was not possible to override an existing entry in the
document store, which can cause issues if the same document is requested
and passed to `find()` twice (The document store would return the old
built document).

By switching to an object we can also make sure that each document is
only present once in the cache.


czosel:
Side note: I implemented this because I thought I'd need it for the navigation feature, but then I found another solution. I could imagine that this bites someone else in the future, so it might be a good idea to merge it already anyway?

PR #115, opened 2019-04-04
Add namespace to forms

### LINKED ISSUE #114: Form builder: Namespace is only applied to questions not to forms
The namespace which was introduced in #97 is just added to the question slug but not the form slug! To make ember-caluma fully namespace aware it should also be added to the form slug.

fkm:
Plus, fixes a problem with the validation.

Resolves #114

### DIFF
--- a/addon/components/cfb-form-editor/general.js
+++ b/addon/components/cfb-form-editor/general.js
@@ -2,6 +2,7 @@ import Component from "@ember/component";
 import { inject as service } from "@ember/service";
 import { computed } from "@ember/object";
 import layout from "../../templates/components/cfb-form-editor/general";
+import { settled } from "@ember/test-helpers";
 import { task, timeout } from "ember-concurrency";
 import { ComponentQueryManager } from "ember-apollo-client";
 import validations from "../../validations/form";
@@ -51,6 +52,7 @@ export default Component.extend(ComponentQueryManager, {
     this._super(...arguments);
 
     await this.get("data").perform();
+    await settled();
 
     if (!this.get("slug")) {
       this.addSlug();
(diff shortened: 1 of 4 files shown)


addon/components/cfb-form-editor/general.js
```diff
@@ -20,11 +21,40 @@ export default Component.extend(ComponentQueryManager, {
...
+      const parent = input.parentElement;
+
+      Object.assign(span, {
+        className: "slugnamespace-slug",
+        innerHTML: `${this.namespace}-`
+      });
+      parent.classList.add("slugnamespace");
+      parent.insertBefore(span, input);
+    }
+  },
```
czosel: Is this _exactly_ the same as for questions? If yes, it might be worth extracting this into a shared function?

fkm: It is. But I'd rather concentrate my efforts on writing the custom component.

czosel: Alright!

fkm:
Coverage is lacking. Somehow the component is still in its loading state when the `addSlug` method is invoked. This should not be the case as I changed the `didReceiveAttrs` lifecycle hook to `async` and added an `await` to the `.perform()`.

https://github.com/projectcaluma/ember-caluma/pull/115/files#diff-68e6340d2cf7a6ed2819d387bc84e37dR50

Works outside of testing and works in testing in `question.js` where I did the same thing. As I intend to write a custom component for the prefixed slug, I recommend to ignore the missing coverage in this case.

fkm:
As this only fails in `ember-canary` and only on some code that will soon be removed by another PR I recommend that we merge this one despite the failed test.

czosel:
Hmm, looks like the test with floating deps is failing, or am I seeing something wrong?

fkm:
Superseded by #116

PR #116, opened 2019-04-04
Add custom component for prefixed slugs

### LINKED ISSUE #113: Option service: Replace dom manipulation with a custom render component
The best way to add the namespace to the input field would probably be to write a render component for the namespace input case. The solution we have now works, so this is just a optimization proposal. ;)

fkm:
There is the problem that it is possible to override an existing form/question if one is faster than the client-side validation. The easiest way to reproduce that is to change the name and then press enter.

That was already the case before. I don't know what the expected behavior is, but I guess it would make sense to make an additional check on the backend.

Fixes #113 

Depends on #125


czosel:
Since `slug` is the primary key of the question, the backend can't tell the difference between two questions with the same slug. We have to make sure that the frontend doesn't issue `SaveQuestion` mutations before checking for slug collisions.

czosel:
Can you rebase/merge master to fix the tests?

addon/components/cfb-form-editor/question.js
```diff
@@ -293,6 +268,12 @@ export default Component.extend(ComponentQueryManager, {
 
       optional([this.get("on-after-submit")])(question);
     } catch (e) {
+      const slug = changeset.get("slug");
+      const prefix = `${this.namespace}-`;
+      if (slug.startsWith(prefix)) {
+        changeset.set("slug", slug.replace(prefix, ""));
+      }
+
```
czosel: Why is this part of the catch block? What's triggering the error?

addon/components/cfb-form-editor/question.js
```diff
@@ -173,6 +145,9 @@ export default Component.extend(ComponentQueryManager, {
   namespace: computed("calumaOptions._namespace", function() {
     return slugify(this.calumaOptions.getNamespace() || "") || null;
   }),
+  prefix: computed("namespace", function() {
+    return this.namespace ? `${this.namespace}-` : "";
+  }),
```
czosel: Might be possible to replace some of the `${this.namespace}-` calls further down in the file with `this.get("prefix")`

fkm:
Use ember-validated-form input component as block.

czosel:
@fkm Is this still WIP?

fkm:
All good :smiley:

addon/components/cfb-form-editor/question.js
czosel: Ah, so you're resetting the change made earlier. I just pushed a commit that assigns the prefixed slug to a local variable instead of to the changeset. This has two advantages:

* no need to reset the value in case the save operation fails
* no visible "jump" from non-prefixed to prefixed slug in the user interface

fkm: Nice! :sunglasses:

PR #117, opened 2019-04-05
UPDATE dependency to upstream master

fkm:
Just until a version with the necessary changes has been released.


czosel:
Maybe better to pin it to the specific commit?

fkm:
Good idea!

PR #137, opened 2019-04-17
WIP: Add file questions

fkm:
There are some things missing:

* [x] Tests
* [ ] Validation
* [ ] Fake answer.value if undefined
* [ ] Better mock for [GenericScalar](https://github.com/graphql-python/graphene/blob/master/graphene/types/generic.py)

### DIFF
--- a/addon/components/cf-field/input/file.js
+++ b/addon/components/cf-field/input/file.js
@@ -14,6 +14,7 @@ export default Component.extend({
   apollo: service(),
 
   downloadUrl: reads("field.answer.fileValue.downloadUrl"),
+  downloadName: reads("field.answer.fileValue.metadata.object_name"),
 
   placeholder: computed("field.answer.fileValue", function() {
     return this.get("field.answer.fileValue")
@@ -54,6 +55,15 @@ export default Component.extend({
       request.send(file);
       request.onload = event => {
         if (event.target.status == 200) {
+          this.set(
+            "field.answer.fileValue.downloadUrl",
+            saved.fileValue.downloadUrl
+          );
+          this.set("field.answer.fileValue.metadata", {
+            object_name: file.name
+          });
+
+          target.value = "";
           target.parentNode.querySelector("[type=text]").value = "";
         }
       };
--- a/addon/components/cf-field/input/file.js
+++ b/addon/components/cf-field/input/file.js
@@ -6,6 +6,35 @@ import { inject as service } from "@ember/service";
 import getFileAnswerInfoQuery from "ember-caluma/gql/queries/get-fileanswer-info";
 import layout from "../../../templates/components/cf-field/input/file";
 
+/**
+ * Promise wrapper around XHMLHttpRequest
+ *
+ * @param {File} file The file to upload.
+ * @param {String} url The MinIO uploadUrl
+ * @return {Promise<Event>} A promise resolving or rejecting with the event.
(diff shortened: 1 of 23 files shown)


addon/components/cf-field/input/file.js
```diff
@@ -0,0 +1,64 @@
...
+    async save({ target }) {
+      const file = target.files[0];
+
+      if (!file) {
+        return;
+      }
+
+      const saved = await this.onSave(file.name);
+
+      const data = new FormData();
```
czosel: Just to be sure: This works in IE11, right? ([caniuse](https://caniuse.com/#feat=xhr2) just mentions limited support for JSON content types)

fkm: This shouldn't be a problem as we append a file. But I'll check anyway.

fkm:
There's a ticket for the missing dot:
https://github.com/projectcaluma/caluma/issues/387

I will look for a solution to the missing metadata with @open-dynaMIX tomorrow.

addon/components/cf-field/input/file.js
fkm: I haven't gotten to test the FormData. IE 11 breaks long before that on a `class` statement and an arrow function. Probably needs some modification to the build process to transform the things IE doesn't understand. But I suggest we open another ticket for that as there are many more issues with IE. (i.e. I remember seeing errors about `Object` missing `assign` from a previous test.

czosel: Oh, good to know! This sounds like something we should look at with high priority then.

fkm:
@czosel Das Ding ist nun soweit fertig.

Die Tests sind noch nicht wirklich fertig. Aber Coverage sollte anständig sein und alles funktioniert.

czosel:
@fkm can you bring the tests in a state that they're passing? (if needed by skipping some?)

addon/gql/queries/get-fileanswer-info.graphql
```diff
@@ -0,0 +1,14 @@
+query getFileAnswerInfo($id: ID!) {
+  node(id: $id) {
+    id
+
```
czosel: nitpick: whitespace

fkm: Whitespace? The empty line after `id`?

czosel: yep - or is this something we do anywhere else already?

fkm: I'll have to check. It is something I do elsewhere :smile:

fkm: We don't. Change is coming with the next commit.

addon/components/cf-field/input/file.js
```diff
@@ -6,6 +6,35 @@ import { inject as service } from "@ember/service";
...
 import layout from "../../../templates/components/cf-field/input/file";
 
+/**
+ * Promise wrapper around XHMLHttpRequest
+ *
+ * @param {File} file The file to upload.
+ * @param {String} url The MinIO uploadUrl
+ * @return {Promise<Event>} A promise resolving or rejecting with the event.
+ */
+function uploadFile(file, url) {
```
czosel: :+1: I was tempted to mention this in review as well :wink:

PR #149, opened 2019-04-19
Renovate: Use semver ranges for dependencies

czosel:
See
https://renovatebot.com/docs/dependency-pinning/#ranges-for-libraries


tests/dummy/app/routes/index.js
```diff
@@ -40,7 +40,7 @@ export default Route.extend(RouteQueryManager, {
       {
         query: gql`
           query {
-            allDocuments {
+            allDocuments(form: "main") {
```
fkm: Is "main" the default slug for `server.create("form")`?

czosel: oops, that was not supposed to be in there - I just changed this locally for testing.

czosel: Fixed.

PR #154, opened 2019-04-24
feat(jexl): add cross-form jexl resolution

czosel:
This allows specifying cross-form jexl expressions like that:

- "sub1.question1"|answer // will look for answer to "question1" in child form "sub1"
- "parent.question1"|answer // will look for answer to "question1" in parent document
- "parent.sub1question1"|answer // will look answer to "question1" in sibiling document "sub1"


czosel:
@kaldras This is not completely finished yet, but I'd like to get your feedback on the design.

kaldras:
As discussed with @czosel this seems like a good solution! :+1: If the backend implementation proves no problem or issues which we didn't thought about, I suppose we implement it this way.

czosel:
@kaldras @fkm ready for review.

fkm:
I wonder if we could have a problem with people naming their forms/documents "parent".

But I like the idea of the traversal! Close enough to DOM to feel like home :smiley:

czosel:
I had the same thought yesterday! What do you think about making sure that form's cant be called `parent` in the form builder?

fkm:
Yeah. We could introduce a list of reserved words. I would have just called it "PARENT" or something for the traversal as slugs need to be lower-case. But I like your idea! Much cleaner.

czosel:
I've also toyed with the idea of using `../` to get one layer up, but I decided against it because then people will start expecting all the other features of file system traversal like `./` and whatnot :wink:

fkm:
I also prefer the named route. Having reserved words for slugs might come in handy later on anyway. And if it's a list of words we only need one error-message and other handling.

czosel:
@fkm I tweaked the API as suggested in https://github.com/projectcaluma/caluma/pull/398#discussion_r279346810.
Can you take another look?

PR #159, opened 2019-04-26
ADD uk-drop options

fkm:


addon/templates/components/cf-field/info.hbs
```diff
@@ -1,6 +1,6 @@
 <button type="button" class="uk-icon-button" uk-icon="info"></button>
 
-<div uk-drop>
+<div uk-drop="pos: bottom-justify; boundary: .uk-form-controls; boundary-align: true">
```
velrest: Is this a breaking change for other implementations?

fkm: Nope. This just changes the visual appearance.

PR #173, opened 2019-04-29
Update Caluma backend and GraphQL schema + types

fkm:


fkm:
Coverage is probably an error from the service. I don't see where the test coverage could be lacking more than before.

PR #179, opened 2019-05-01
Add capability to remove existing answers

fkm:


addon/lib/field.js
```diff
@@ -178,20 +180,32 @@ export default EmberObject.extend(Evented, {
...
-        mutation: this.get(`saveDocument${type}Mutation`),
+    if (value === null || value.length === 0) {
+      return yield this.apollo.mutate({
+        mutation: removeAnswerMutation,
         variables: {
           input: {
-            question: this.get("question.slug"),
-            document: this.get("document.id"),
-            value: this.get("answer.value")
+            answer: atob(this.get("answer.id")).split(":")[1]
```
czosel: Use `import { atob } from "ember-caluma/helpers/atob";` instead

docker-compose.yml
```diff
@@ -34,4 +34,4 @@ services:
 
 volumes:
   dbdata:
-  minio_data:
\ No newline at end of file
+  minio_data:
```
czosel: Can you configure your editor to stop doing that? ^^

tests/integration/components/cf-field-test.js
```diff
@@ -46,6 +46,16 @@ module("Integration | Component | cf-field", function(hooks) {
...
   });
 
+  test("it allows deleting existing input", async function(assert) {
+    assert.expect(1);
+
+    await render(hbs`{{cf-field field=field}}`);
+    this.set("field.answer.value", "Test");
+
+    await fillIn("input", "");
+    assert.equal(this.field.answer.value, "", "Value was removed.");
```
czosel: Shouldn't the whole answer be removed?

docker-compose.yml
fkm: Newlines at EOF?

addon/lib/field.js
fkm: Ok

What's the reason behind the helper?

tests/integration/components/cf-field-test.js
fkm: Good point! I'll look into that. Probably need to add Mirage to the test.

docker-compose.yml
czosel: yep

addon/lib/field.js
czosel: The helper just exists to avoid having the same logic spread across many files.

docker-compose.yml
fkm: That's something I mostly do manually. But yeah. I can stop.

What's the reasoning behind not having newlines at the end of the files for some file types? And what types require newlines and which don't?

czosel: Ah ok. I don't know why it's preferred not to have those newlines, but I do know that it's nice to have a consistent default for this all over the repository. My editor is configured to remove those newlines, and I think most of the people in our team handle it the same way.
I'd just like to get of the extra noise in those merge requests :wink:

fkm: Ok

tests/integration/components/cf-field-test.js
fkm: I don't really see a way to create the fields from (Mirage)GraphQL without the cf-form component. I'll have another look at it later. This is probably going to take a moment.

czosel: Wanna create an issue and move on for the moment?

fkm: Maybe the best solution to get stuff done quickly.

czosel: Alright, please don't forget to create the issue :wink:

fkm: https://github.com/projectcaluma/ember-caluma/issues/181

PR #187, opened 2019-05-02
fix(table): render download links for file questions in tables

czosel:
I left out the extra query to the caluma backend for the moment. We can add it later if it turns out to be needed.

### DIFF
--- a/addon/components/cf-field-value.js
+++ b/addon/components/cf-field-value.js
@@ -1,13 +1,17 @@
 import Component from "@ember/component";
 import layout from "../templates/components/cf-field-value";
 import { computed } from "@ember/object";
+import { inject as service } from "@ember/service";
+import getFileAnswerInfoQuery from "ember-caluma/gql/queries/get-fileanswer-info";
 
 export default Component.extend({
   layout,
 
+  apollo: service(),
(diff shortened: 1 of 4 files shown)


addon/components/cf-field-value.js
```diff
@@ -1,34 +1,63 @@
...
         const answerValue = field.answer.value || [];
         const options = field.question.multipleChoiceOptions.edges.filter(
           edge => answerValue.includes(edge.node.slug)
         );
-        return options && options.length
-          ? options.map(edge => edge.node.label).join(", ")
-          : answerValue.join(", ");
+        return {
+          label:
+            options && options.length
```
fkm: Is it possible that there are no options but a non-empty answerValue?

czosel: Not sure. I didn't change anything here, this was just prettier doing some reformatting. Let's leave it for the moment :wink:

fkm: Okey :-)

Plus. I just realized that the chosen options could have been removed from the question.

czosel: :+1: Is this good to go, then?

PR #189, opened 2019-05-02
Add "root" to doc slug lookups, improve reporting

winged:
New: When evaluating JEXL, you can also use `root` as a path
element for your slug, which will then start at the top-level
document.

Also, improve error reporting if a path cannot be resolved:
We can now report exactly the position where it failed

Note this is the equivalent of projectcaluma/caluma#414

### DIFF
--- a/addon/lib/document.js
+++ b/addon/lib/document.js
@@ -86,25 +86,57 @@ export default EmberObject.extend({
     const segments = slugWithPath.split(".");
     const slug = segments.pop();
     const doc = this.resolveDocument(segments);
-    return doc && doc.fields.find(field => field.question.slug === slug);
+    let field = doc && doc.fields.find(field => field.question.slug === slug);
+    if (!field) {
+      segments.push(slug);
+      this._resolveError(segments, slug, doc);
+    }
+    return field;
   },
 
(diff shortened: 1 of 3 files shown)


addon/lib/document.js
```diff
@@ -86,25 +86,57 @@ export default EmberObject.extend({
     const segments = slugWithPath.split(".");
     const slug = segments.pop();
     const doc = this.resolveDocument(segments);
-    return doc && doc.fields.find(field => field.question.slug === slug);
+    let answer = doc && doc.fields.find(field => field.question.slug === slug);
```
czosel: This is actually a `field`, not an `answer`. The client side model layer is structured like this:
```
document --(1:n)--> field --(1:1)--> question
                          \---(1:0)--> answer
```

addon/lib/document.js
```diff
@@ -86,25 +86,57 @@ export default EmberObject.extend({
...
+    if (answer === undefined) {
```
czosel: Since you're checking for object presence
```suggestion
    if (!answer) {
```

winged: Fixed

PR #195, opened 2019-05-03
fix(deps): upgrade to babel 7

czosel:


czosel:
This still need to be tested with a consuming app, and the according release notes have to be added (I'm expecting this to be a breaking change)

fkm:
Breaking because of the Ember CLI update?

czosel:
I expect this to be breaking because probably we'll have to upgrade the babel config in consuming apps as well.

fkm:
Ok

czosel:
Just tested it, and it turns out that this isn't breaking at all! :+1:

PR #199, opened 2019-05-07
fix(ie 11): add "manual" polyfill for array.flat

czosel:


czosel:
This was just a quick hack for IE 11 compat of the form. We should only merge it if we don't find out why babel didn't polyfill it.

addon/components/cf-field/input/table.js
```diff
@@ -97,6 +102,8 @@ export default Component.extend(ComponentQueryManager, {
 
         this.set("showModal", false);
       } catch (e) {
+        // eslint-disable-next-line no-console
+        console.error(e);
```
fkm: Do we really need log the error when we already show a message?

addon/components/cf-field/input/table.js
```diff
@@ -6,6 +6,11 @@ import saveDocumentTableAnswerMutation from "ember-caluma/gql/mutations/save-doc
 import { inject as service } from "@ember/service";
 import { ComponentQueryManager } from "ember-apollo-client";
 
+// TODO check why babel polyfill didn't pick this up
```
fkm: I would recommend adding a real polyfill.

```js
if (!Array.prototype.flatMap) {
  Array.prototype.flatMap = function() {
    return Array.prototype.map.apply(this, arguments).flat(1);
  };
}
```

[Source](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#Polyfill)

addon/components/cf-field/input/table.js
czosel: This is just supposed to help with analyzing why the generic error message shows up. Ideally, messages like tihs should be sent to a remote service (sentry?), but as long as we don't have anything like that a message in the console seems like the best possible alternative to me. 
To me, it's not only helpful in development, but also when users see the error in production and we're trying to analyze remotely.

addon/components/cf-field/input/table.js
czosel: I think the best would be to make sure that babel polyfills this. [This post](https://stackoverflow.com/a/55420761) mentions that @babel/preset-env should in theory be able to do it. We might have to merge #195 first.

addon/components/cf-field/input/table.js
fkm: Ok. I can accept that :smiley:

addon/components/cf-field/input/table.js
fkm: True. But you will have to update your code as well when you do as the syntax is different. But yeah. As you will have to remove the polyfill you will have to touch the code anyway.

addon/components/cf-field/input/table.js
anehx: You could also use `warn()` of `@ember/debug` to hide it in production builds since we don't want to spam the users console, do we?

addon/components/cf-field/input/table.js
anehx: If I understand [this issue](https://github.com/babel/babel/issues/9749) correctly we need core-js@3 which only works with babel@7 AFAIK. So we should try to merge #195 before we add some dirty solution which this definitely is..

anehx: If I understand [this issue](https://github.com/babel/babel/issues/9749) correctly we need core-js@3 which only works with babel@7 AFAIK. So we should try to merge #195 before we add some dirty solution which this definitely is..

czosel: @fkm I just pushed this PR to illustrate the problem - if we can get the polyfill for this to work in the next couple of days, we can close this PR without ever merging it.

PR #203, opened 2019-05-07
Scroll to top on document change

fkm:
Not sure if `finally` is the right place.

The reasoning is that I cannot add it at the end of the `try` block as there are `return`s before that. My problem with `finally` is that it will scroll on error which might be confusing.

### DIFF
--- a/addon/components/cf-navigation.js
+++ b/addon/components/cf-navigation.js
@@ -67,6 +67,8 @@ export default Component.extend(ComponentQueryManager, {
         // eslint-disable-next-line no-console
         console.error(e);
         return null;
+      } finally {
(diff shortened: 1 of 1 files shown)


czosel:
Do you know what kind of exceptions we're expecting in that try/catch block?

fkm:
No. As there is no API call I guess it's about the data structure.

czosel:
Might be a good idea to refactor the code with things like `getWithDefault` and similar safety checks?

fkm:
That's a task for someone who actually knows what's going on (data structures and requirements). I didn't even know that this component existed until today.

czosel:
Probably, that would be me :sweat_smile:

PR #207, opened 2019-05-07
Add setting to open links in new window

fkm:


addon/templates/components/cf-field/info.hbs
```diff
@@ -1,5 +1,5 @@
 <button type="button" class="uk-icon-button" uk-icon="info" {{action (mut showModal) true}}></button>
 
 {{#uk-modal visible=showModal}}
-  {{markdown-to-html text}}
+  {{markdown-to-html text openLinksInNewWindow=true}}
```
czosel: Should we add the same option to the "static text" component?

fkm: Good point!

fkm: Done

PR #209, opened 2019-05-08
chore(release): add travis config for semantic release

czosel:


czosel:
Ah damn, copy-paste error. The branch is supposed to be called "release".

PR #218, opened 2019-05-08
feat(cf-form): allow passing context information to cf-form, cf-navigation

czosel:
This is useful to allow passing "context" data to custom components.

### DIFF
--- a/addon/components/cf-navigation.js
+++ b/addon/components/cf-navigation.js
@@ -29,7 +29,7 @@ export default Component.extend(ComponentQueryManager, {
    * Can be used to pass "context" information from the outside through
    * to custom overrides.
    *
-   * @argument {*} overrides
+   * @argument {*} context
    */
   context: null,
 
--- a/addon/components/cf-navigation.js
+++ b/addon/components/cf-navigation.js
@@ -6,12 +6,46 @@ import { ComponentQueryManager } from "ember-apollo-client";
 import { task } from "ember-concurrency";
 import getNavigationQuery from "ember-caluma/gql/queries/get-navigation";
 
+/**
+ * Component to display a nested form including navigation.
+ *
+ * ```hbs
+ * {{cf-navigation documentId="myDocId" section=section subSection=subSection}}
+ * ```
+ *
+ * @class CfFormComponent
+ */
 export default Component.extend(ComponentQueryManager, {
   layout,
   documentStore: service(),
 
+  /**
+   * The ID of the nested document to display the navigation for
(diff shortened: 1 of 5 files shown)


addon/components/cf-navigation.js
```diff
@@ -6,12 +6,46 @@ import { ComponentQueryManager } from "ember-apollo-client";
...
 export default Component.extend(ComponentQueryManager, {
   layout,
   documentStore: service(),
 
+  /**
+   * The ID of the nested document to display the navigation for
+   * @argument {String} documentId
+   */
   documentId: null,
-  activeDocumentId: null,
```
anehx: Will this be a breaking change?

addon/components/cf-navigation.js
```diff
@@ -6,12 +6,46 @@ import { ComponentQueryManager } from "ember-apollo-client";
...
+
+  /**
+   * Can be used to pass "context" information from the outside through
+   * to custom overrides.
+   *
+   * @argument {*} context
```
anehx: Maybe we should explicitly expect an object..

addon/components/cf-navigation.js
```diff
@@ -6,12 +6,46 @@ import { ComponentQueryManager } from "ember-apollo-client";
...
+   * @argument {String} section
+   */
+  section: null,
+
+  /**
+   * Form slug of currently visible sub-section
+   *
+   * @argument {String} subSection
+   */
+  subSection: null,
```
anehx: Where is `section` and `subSection` used? Is this really something that needs to be in caluma or is it customer specific implementation?

addon/components/cf-navigation.js
anehx: Also, what is this even? :joy:

czosel: This is just a left-over from a previous implementation - I just cleaned it up :wink:

addon/components/cf-navigation.js
czosel: Is there any advantage for limiting the accepted types? We're just passing the context further and don't really care about it's datatype or it's content.

addon/components/cf-navigation.js
czosel: Those properties are tracking which sub-form is currently active. Since query params are handled in controllers, this lives currently outside of ember-caluma. We should add an usage example for such a controller in another PR.

czosel:
@anehx We don't have an existing test case for componentOverrides yet - and I don't have the time to create one right now :unamused:

PR #220, opened 2019-05-09
fix: consider empty but required fields

fkm:
Not sure if that fixes everything. But it would at least be a start :smile:

### DIFF
--- a/addon/lib/document.js
+++ b/addon/lib/document.js
@@ -185,7 +185,14 @@ export default EmberObject.extend({
       }
 
       const visibleFields = this.fields.filter(f => !f.question.hidden);
-      if (visibleFields.every(f => f.isValid)) {
+      const requiredFields = visibleFields.filter(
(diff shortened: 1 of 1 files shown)


addon/lib/document.js
```diff
@@ -185,7 +185,14 @@ export default EmberObject.extend({
       }
 
       const visibleFields = this.fields.filter(f => !f.question.hidden);
-      if (visibleFields.every(f => f.isValid)) {
+      const requiredFields = visibleFields.filter(
+        f => f.question.isRequired !== "false"
```
anehx: Isn't `question.isRequired` a JEXL expression?

fkm: From it being a string; that's possible. Let me check if I can find the right property.

czosel: As discussed, should be `optional`

fkm: Done

PR #228, opened 2019-05-13
 fix(validation): add validation for static, fix multiple choice

czosel:
Depends on #225

### DIFF
--- a/addon/components/cf-field/input/table.js
+++ b/addon/components/cf-field/input/table.js
@@ -107,6 +107,8 @@ export default Component.extend(ComponentQueryManager, {
 
       this.set("showModal", false);
     } catch (e) {
+      // eslint-disable-next-line no-console
+      console.error(e);
       this.get("notification").danger(
         this.get("intl").t("caluma.form.notification.table.add.error")
       );
--- a/addon/lib/field.js
+++ b/addon/lib/field.js
@@ -235,13 +235,11 @@ export default EmberObject.extend(Evented, {
    */
   validate: task(function*() {
     const specificValidation = this.get(`_validate${this.question.__typename}`);
-    if (!specificValidation) {
-      // eslint-disable-next-line no-console
-      console.error(
-        "Missing validation function for",
-        this.question.__typename
-      );
-    }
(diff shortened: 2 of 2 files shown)


addon/components/cf-field/input/table.js
```diff
@@ -107,6 +107,8 @@ export default Component.extend(ComponentQueryManager, {
 
       this.set("showModal", false);
     } catch (e) {
+      // eslint-disable-next-line no-console
+      console.error(e);
```
anehx: Leftover I guess?

addon/lib/field.js
```diff
@@ -234,9 +234,17 @@ export default EmberObject.extend(Evented, {
...
    */
   validate: task(function*() {
+    const specificValidation = this.get(`_validate${this.question.__typename}`);
+    if (!specificValidation) {
+      // eslint-disable-next-line no-console
+      console.error(
+        "Missing validation function for",
+        this.question.__typename
+      );
+    }
```
anehx: Use `assert` from `@ember/debug` for those kind of logs / errors

addon/components/cf-field/input/table.js
czosel: No, this is actually intended to avoid swallowing exceptions.

addon/lib/field.js
czosel: Fixed. Still have to figure out why the test is failing ...

czosel: Found it, was a curly brace ...

addon/components/cf-field/input/table.js
anehx: We used this for debugging the `Array.flat`-IE11-affair..

fkm: Let's keep the error logs until we have fully deployed ember-caluma in all our (main) projects and have more time to harden and clean-up the code.

czosel: @anehx True, but I still like to have those statements around. Lets discuss in person (ETA in office 30min :wink: )

PR #230, opened 2019-05-13
chore: update ember-uikit

fkm:


yarn.lock
```diff
@@ -7237,7 +7237,7 @@ ember-try@1.1.0:
     rsvp "^4.7.0"
     walk-sync "^0.3.3"
 
-ember-uikit@^0.8.0:
+ember-uikit@^0.8.1:
   version "0.8.1"
```
czosel: Looks like 0.8.1 was installed already

PR #233, opened 2019-05-13
fix: remove jexl logic and hide toggle

fkm:


addon/templates/components/cfb-form-editor/question.hbs
```diff
@@ -76,7 +76,7 @@
           {{/if}}
         </div>
 
-        {{#if (not-eq f.model.__typename "StaticQuestion")}}
+        {{#if (and (or (eq f.model.isRequired "true") (eq f.model.isRequired "false")) (not-eq f.model.__typename "StaticQuestion"))}}
```
czosel: I'd extract this into the `js` file.

fkm: I agree in principle. But as all the other display logic is in the template, I did so as well.

czosel: Yeah, I think the "soft limit" for this kind of logic should be 2-3 assertions. Would you mind extracting it?

fkm: "Soft limit"?

Anyways. Is there a good way to observe a Changeset in a computed property?

czosel: Looks like you found out :+1:

fkm: Nope. This doesn't work.

czosel: maybe @anehx can help?

addon/components/cfb-form-editor/question.js
```diff
@@ -189,6 +189,13 @@ export default Component.extend(ComponentQueryManager, {
     return namespace ? `${namespace}-` : "";
   }),
 
+  requiredIsVisible: computed("changeset.{__typename,isRequired}", function() {
+    return (
+      (this.model.isRequired === "true" || this.model.isRequired === "false") &&
```
czosel: Shouldn't this be `this.changeset`?

fkm: Indeed. And it solves the problem.

The `this.model` was from my first attempt and I didn't test but went directly to Google. Thank you!

Issue #236, opened 2019-05-14
Table rendering: Make displayed columns configurable


czosel:
It would be great if the displayed columns in a table were configurable in the form builder (currently we're always rendering the first 4 columns). For storage, I'd suggest using `meta` on the TableQuestion. The most basic implementation I can think of is just a static list of question slugs. This list could potentially run out-of-sync with the form backing the table, which something I wouldn't solve in a first version.

A more advanced version would allow for dynamic column contents, e.g. two field values joined together in one column (might be possible using JEXL). But that seems like something that we better leave out for the moment.

For the basic version, I'd expect a checkbox group with all the table's questions in the form builder UI.

/cc @fkm @kaldras @anehx

anehx:
:+1:

czosel:
Implemented in #237 
/cc @JohnL17

PR #237, opened 2019-05-14
feat: add table column display configuration

Yelinz:
Basic implementation of #236

### DIFF
--- a/addon/components/cfb-form-editor/question.js
+++ b/addon/components/cfb-form-editor/question.js
@@ -431,7 +431,11 @@ export default Component.extend(ComponentQueryManager, {
       this.get("validateSlug").perform(this.prefix + value, changeset);
     },
 
-    selectColumnToDisplay(value) {
+    /*
+     * This function adds the selected slugs to the selectedColumnsToDisplay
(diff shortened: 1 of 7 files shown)


addon/components/cfb-form-editor/question.js
```diff
@@ -420,6 +429,17 @@ export default Component.extend(ComponentQueryManager, {
...
+
+    selectColumnToDisplay(value) {
+      let arr = this.get("model.selectedColumnsToDisplay");
+      let idx = arr.indexOf(value);
+      if (idx >= 0) {
+        arr.splice(idx, 1);
+        this.set("model.selectedColumnsToDisplay", arr);
+      } else {
+        this.set("model.selectedColumnsToDisplay", [...arr, value]);
+      }
```
czosel: Took me a moment to figure out what you're doing here - maybe `toggleColumnToDisplay` would be a better name? And/or add a comment explaining that if the function receives a slug that is already select, it will deselect it?

Yelinz: Renamed and added a comment

PR #238, opened 2019-05-14
feat(table): remove action buttons in disabled state

Yelinz:


addon/templates/components/cf-field/input/table.hbs
```diff
@@ -22,20 +22,24 @@
...
-      <tr>
-        <td colspan={{add (count-keys questions) 1}} class="uk-text-center">
-          {{#uk-button size="small" on-click=(perform addRow)}}{{uk-icon "plus"}}{{/uk-button}}
-        </td>
-      </tr>
-    </tfoot>
+    {{#unless disabled}}
+      <tfoot>
+        <tr>
+          <td colspan={{add (count-keys questions) 1}} class="uk-text-center">
```
fkm: The count is no longer valid if the column with the buttons is not shown.

This will probably become ugly when you add the necessary logic. But we need something to avoid (sort of) broken table layouts.

fkm: The same goes for the empty TH on line 8.

I mean the removal when disabled. The logic will be nicer.

addon/templates/components/cf-field/input/table.hbs
```diff
@@ -22,20 +22,24 @@
               <td>{{cf-field-value field=field}}</td>
             {{/each}}
           {{/if}}
-          <td class="uk-text-right">
-            <button type="button" data-test-edit-row class="uk-icon-button" uk-icon="pencil" {{action "editRow" document}}></button>
-            <button type="button" data-test-delete-row class="uk-icon-button" uk-icon="trash" {{action (perform deleteRow) document}}></button>
-          </td>
+          {{#unless disabled}}
```
fkm: As discussed. Just put the unless around one of the buttons and you are fine.

PR #243, opened 2019-05-15
fix: add DDAU handlers

fkm:


addon/templates/components/cf-field/input/table.hbs
```diff
@@ -29,7 +29,12 @@
       </tr>
     </tfoot>
   </table>
-  {{#uk-modal visible=(and showModal documentToEdit) bgClose=false}}
+  {{#uk-modal
+    visible=(and showModal documentToEdit)
+    bgClose=false
+    on-show=(action (mut showModal) true)
```
czosel: This is not required, right?

fkm: I removed `on-show`.

I kept back the dependency via commit hash. Either we merge this into the master and update the dependency when we have release the commit or we keep this PR open until ember-uikit is ready.

IMO this PR should not break the master as it will just provide another handler that will not be used.

czosel: I agree, we can merge this already anyway.

czosel:
Can you still change the commit message into something that gives more context when displayed in the changelog? Maybe something like
fix(table): fix bug where Dialog doesn’t open after closing ?

fkm:
Rebasing somehow didn't work.
> error: invalid line 1: fix(table): fix bug where dialog doesn't open after closing

So I reset and recommitted. I hope this doesn't have some side effect.

PR #255, opened 2019-05-18
feat(cf-navigation): add possibility to pass custom headers

czosel:


czosel:
This is just a workaround, which I prefer to leave undocumented for the moment - we need to find a proper solution for this in the long run.

PR #256, opened 2019-05-20
fix(navigation): fix document states in navigation

anehx:


addon/lib/document.js
```diff
@@ -170,52 +180,60 @@ export default EmberObject.extend({
...
-        : "unfinished";
+      return "unfinished";
     }
   ),
 
   state: computed("childState", "ownState", function() {
-    return STATE_PRECEDENCE.find(state =>
-      [this.get("childState"), this.get("ownState")].includes(state)
-    );
+    return getParentState([this.childState, this.ownState].filter(Boolean));
```
fkm: Can childState or ownState be empty?

anehx: Yes, if they do not have any fields they will be `null`

PR #261, opened 2019-05-21
feat(form): enable widget overrides for forms

anehx:


addon/templates/components/cf-field.hbs
```diff
@@ -4,11 +4,7 @@
   {{/if}}
   <div class="uk-flex">
     <div class="uk-width-expand">
-      {{#if componentOverride}}
-        {{component componentOverride field=field disabled=disabled context=context onSave=(perform save)}}
-      {{else}}
-        {{cf-field/input field=field disabled=disabled onSave=(perform save)}}
-      {{/if}}
+      {{component (get-widget field) field=field disabled=disabled context=context onSave=(perform save)}}
```
fkm: I wonder if it would be better to set `default="cf-field/input"` here instead of having it 'hidden' in the helper. IMO it would help with the readability of the code.

PR #262, opened 2019-05-21
fix: do not display warning if no override

fkm:


addon/helpers/get-widget.js
```diff
@@ -25,6 +25,11 @@ export default Helper.extend({
   compute([field], { default: defaultWidget = "cf-field/input" }) {
     try {
       const widget = field.question.meta.widgetOverride;
+
+      if (!widget) {
+        return defaultWidget;
+      }
```
anehx: Would be better if you used `warn("bla bla", widget, { id: "bla-bla" })`

See [here](http://api.emberjs.com/ember/3.9/functions/@ember%2Fdebug/warn)

fkm: Hm?

anehx: Gotta remove those lines as well tho

fkm: :joy:

fkm: Done

PR #268, opened 2019-05-22
chore(ci): run tests against the last LTS ember version

anehx:


.travis.yml
```diff
@@ -43,9 +50,10 @@ jobs:
     # we recommend new addons test the current and previous LTS
     # as well as latest stable release (bonus points to beta/canary)
     - stage: "Additional Tests"
-      env: EMBER_TRY_SCENARIO=ember-release
+      env: EMBER_TRY_SCENARIO=ember-lts-3.4
+    - env: EMBER_TRY_SCENARIO=ember-release
```
fkm: Is this minus sign here correct?

I'm no Yaml specialist and definitely no Travis guru. But if I understand this correctly the ember-release scenario is its own 'stage' (or whatever this is called).

anehx: No this is absolutely correct.. This creates a matrix running "ember-lts-3.4" first and then "ember-release"

PR #271, opened 2019-05-22
Fix various build warnings

anehx:


ember-cli-build.js
```diff
@@ -9,10 +9,7 @@ module.exports = function(defaults) {
     sassOptions: { implementation: sass },
     snippetPaths: ["tests/dummy/app/snippets"],
     babel: {
-      plugins: [
-        ["@babel/plugin-proposal-decorators", { legacy: true }],
-        "@babel/plugin-proposal-object-rest-spread"
-      ]
+      plugins: ["@babel/plugin-proposal-object-rest-spread"]
```
fkm: I thought about doing the same thing. But ask Christian if that's ok. IIRC he added this rather recently to make the build work in IE11.

anehx: That's something the host app should do not the addon. `ember-cli-build.js` doesn't affect the host app, only the dummy app where we definitely don't care about IE... /cc @czosel

PR #278, opened 2019-05-23
chore(ember-uikit): update dependency

fkm:
Currently, points to a commit hash. Can be changed on next ember-uikit release.


fkm:
I might have misunderstood. But from what I've gathered, Zosel doesn't like the resolutions in the projects.

anehx:
Well in that case we have a problem.. Because I detest git dependencies in libraries! :angry: :wink: @czosel

czosel:
I think I was just confused about lockfile. I have no problem with `resolutions` :slightly_smiling_face:

PR #290, opened 2019-05-28
feat(navigation): allow customization of the navigation

anehx:
BREAKING CHANGE: this changes the way a form is used in the host
application. Before, the `cf-navigation` component rendered the whole
navigation and the form in it. Now there is a wrapper component
`cf-content` which yields or renders the form and the navigation. This
way the host app is able to customize where to render the navigation and
the form. Also the host app does not need to pass the `section` and
`subSection` query parameters since they are taken directly from the
router.

Before:
```hbs
{{cf-navigation
  documentId=documentId
  context=context
  section=section
  subSection=subSection
}}
```

After:
```hbs
{{cf-content documentId=documentId context=context}}
```


addon/components/cf-content.js
```diff
@@ -0,0 +1,132 @@
...
+   */
+  context: null,
+
+  /**
+   * Form slug of currently visible section
+   *
+   * @argument {String} section
+   * @readonly
+   */
+  section: reads("router.currentRoute.queryParams.section"),
```
fkm: Sexy! :sunglasses:

PR #293, opened 2019-05-29
feat(navigation): add previous / next pagination

### LINKED ISSUE #287: cf-navigation: Add "previous" and "next" buttons for navigation
Just like in the precursor of this project, it would be nice to display "previous" and "next" buttons below each form.

anehx:
Depends on #290 

Closes #287

### DIFF
--- a/addon/components/cf-pagination.js
+++ b/addon/components/cf-pagination.js
@@ -1,0 +1,115 @@
+import Component from "@ember/component";
+import layout from "../templates/components/cf-pagination";
+import { computed } from "@ember/object";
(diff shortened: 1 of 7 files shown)


fkm:
Looks good. Seems to me though that this doesn't include the sections themselves but only the sub-sections. This is not an issue with our current project(s) but from what I've seen it would theoretically be possible to have forms on the sections too.

anehx:
As discussed, we'll merge this PR but we need to discuss this issue. I created #299 to keep track of it..

PR #295, opened 2019-05-31
fix(form): improve initial loading performance

anehx:
This needs https://github.com/projectcaluma/caluma/pull/473 to be merged.


anehx:
@fkm You're right! Changed it..

PR #300, opened 2019-05-31
chore(atob): rename atob helper to decodeId

### LINKED ISSUE #9: Rename `atob` helper for better clarity
From https://github.com/projectcaluma/ember-caluma-utils/issues/125

The `atob` helper does not only call the native `atob` function but also does a splitting of the id. With the current name it is confusing why not just the native function was used. Better rename the helper to `decodeId` or something like this.

anehx:
Resolves #9

### DIFF
--- a/addon/helpers/atob.js
+++ b/addon/helpers/atob.js
@@ -1,15 +1,23 @@
 import { helper } from "@ember/component/helper";
+import { deprecate } from "@ember/application/deprecations";
+import { decodeId } from "ember-caluma/helpers/decode-id";
 
 export function atob(str) {
-  try {
-    return window.atob(str).split(":")[1];
-  } catch (e) {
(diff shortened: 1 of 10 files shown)


addon/helpers/atob.js
```diff
@@ -1,15 +1,23 @@
...
-  } catch (e) {
-    // eslint-disable-next-line no-console
-    console.warn("Attempted to decode", str, "as base64");
-    return str;
-  }
+  deprecate(
+    'DEPRECATED: Using the `atob` helper is deprecated. Instead, use the new `decodeId` helper: `import { decodeId } from "ember-caluma/helpers/decode-id";`.',
+    false,
+    { id: "ember-caluma.atob-in-code", until: "1.0.0" }
+  );
```
czosel: nice!

PR #326, opened 2019-06-12
feat(form): prevent displaying of empty forms

### LINKED ISSUE #322: Automatically navigate to first subform that contains questions
If the main form contains only `FormQuestion`s, it would be nice to automatically redirect to the first subform that actually contains questions that can be filled out.

anehx:
Resolves #322


addon/components/cf-content.js
```diff
@@ -130,19 +131,12 @@ export default Component.extend(ComponentQueryManager, {
       "allForms.edges"
     )).map(({ node }) => node);
 
-    return buildTree(
-      documents.find(doc => doc.id === rootId),
-      documents,
-      forms
+    return this.documentStore.find(
+      buildTree(documents.find(doc => doc.id === rootId), documents, forms)
```
czosel: Is this a fix for the issue I had yesterday?

czosel:
Does this mean we can also remove https://github.com/projectcaluma/ember-caluma/blob/7690d5a5e986f0c168488a3eb4f92b052569d0a9/addon/components/cf-navigation-item.js#L12 ?

addon/components/cf-content.js
anehx: Just refactoring..

anehx:
@czosel Yes, this is removed now! Please redo your review

addon/components/cf-content.js
```diff
@@ -170,9 +183,124 @@ export default Component.extend(ComponentQueryManager, {
...
+      this.displayedDocumentChanged.perform();
+    }
+  ),
+
+  displayedDocumentChanged: task(function*() {
+    if (isDisplayableDocument(this.displayedDocument) || !this.nextSection) {
+      return;
+    }
+
+    yield timeout(100);
```
czosel: Why not just the next runloop?

anehx: Because I want it to run just once and `Ember.debouce` somehow didn't work..

anehx: I gotta try another solution here. This is not as clean as we want it to be..

fkm: AFAIK yielding a timeout is the 'official' way to debounce inside a task.

czosel: If this is "best practice" I'd also be ok with it. Just the number `100` seem artificial.

fkm: Here's the debounce example from the official docs:
http://ember-concurrency.com/docs/examples/autocomplete/

And aren't the debounce values (almost) always arbitrary? I personally use 300ms but that's a value I got by trial and error while developing frontend widgets.

anehx: > a timeout is the 'official' way to debounce inside a task.

`yield timeout(x)` is definitely best practice for debouncing.

> And aren't the debounce values (almost) always arbitrary? I personally use 300ms but that's a value I got by trial and error while developing frontend widgets.

Problem is, this debounce is not a UX thing here. What I was trying to do is waiting for all the `hiddenTask`s to finish. However I have no way of knowing when they are started or finished.

I'll look more into this, debouncing does not seem like the right way to do this..

anehx: Done!

PR #344, opened 2019-06-17
fix(babel): remove obsolete babel config

### LINKED ISSUE #341: Let babel config be the host apps problem
https://github.com/projectcaluma/ember-caluma/blob/144e81c5d9a5404a2d2f91604bc92424099e190c/index.js#L19-L26

I think we should remove this chunk of code in favor of explaining how to configure the host app the correct way. The babel config should be a concern of the host app, not the addon..

anehx:
Resolves #341

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,10 +19,7 @@ module.exports = EngineAddon.extend({
   // see https://github.com/dfreeman/ember-cli-node-assets/issues/11
   // for why this is not wrapped in `options`
   babel: {
-    plugins: [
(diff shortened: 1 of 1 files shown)


czosel:
Isn't this a breaking change? Do we need to add documentation about how to migrate?

anehx:
How is this breaking? We support v3.8+ which does exactly this per default

czosel:
All good, I read your comment in the issue only after writing my comment above :wink:

Issue #306, opened 2019-06-04
Wrong `field` reference in questions


fkm:
It seems that the same field reference is passed to all questions. But this also includes the answer property on the field.

I'm not sure, where exactly this happens but to illustrate the problem here's a screenshot showing the relevant tree / values.

![Screenshot_20190604_152326](https://user-images.githubusercontent.com/100262/58882540-bccd2980-86dc-11e9-9587-01a387ad53db.png)

anehx:
The new structure (#376) does not pass a `field` to the question - therefore this should be fixed.

PR #347, opened 2019-06-18
fix(table): fix answer handling of table questions

anehx:
This allows us to handle table questions like we handle all other types. Updating client state then becomes obsolete since the save response automatically triggers the update of the computed property.

10/10 would merge!

### DIFF
--- a/addon/components/cf-field/input/table.js
+++ b/addon/components/cf-field/input/table.js
@@ -64,12 +64,12 @@ export default Component.extend(ComponentQueryManager, {
   }).drop(),
 
   deleteRow: task(function*(document) {
-    const remainingDocuments = this.get("field.answer.rowDocuments").filter(
+    const remainingDocuments = (this.get("field.answer.value") || []).filter(
       doc => doc.id !== document.id
     );
 
     // update client-side state
-    this.set("field.answer.rowDocuments", remainingDocuments);
+    this.set("field.answer.value", remainingDocuments);
 
     yield this.onSave(remainingDocuments.map(doc => doc.id));
   }),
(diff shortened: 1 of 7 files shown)


addon/lib/answer.js
```diff
@@ -41,7 +44,21 @@ export default EmberObject.extend({
...
+          return (
+            value &&
+            value.map(raw =>
+              Document.create(getOwner(this).ownerInjection(), {
+                raw,
+                parentDocument: this.document
+              })
+            )
+          );
+        }
```
czosel: Isn't this creating a new `Document` every time the getter is called?

czosel:
Fixes #306

addon/lib/field.js
```diff
@@ -102,16 +101,7 @@ export default EmberObject.extend(Evented, {
             question: { slug: this._question.slug },
             [camelize(__typename.replace(/Answer$/, "Value"))]: null
           },
-          { document: this.document, field: this },
-          this._answer && Array.isArray(this._answer.value)
```
fkm: Would this still allow to get the case state (valid, etc.) after building the document tree?

Meaning, will we still catch invalid inputs in table rows?

addon/components/cf-field/input/table.js
```diff
@@ -82,30 +83,21 @@ export default Component.extend(ComponentQueryManager, {
         return;
       }
 
-      const rows = this.getWithDefault("field.answer.rowDocuments", []);
+      const rows = this.getWithDefault("field.answer.value", []);
```
fkm: This leaves me with `rows === null` :slightly_frowning_face:

czosel:
@fkm Should work now :slightly_smiling_face:

PR #353, opened 2019-06-19
fix(document): add fallback if no fields

fkm:
doc.getWithDefault() would be nice here. But I couldn't reproduce the 
error and didn't want to spend too much time on that.


fkm:
I actually cannot. I had this error after finishing a case. But I didn't catch where exactly that came from and couldn't reproduce it.

fkm:
As - AFAIK - today's test has already started I'll keep this one open and see if I can reproduce the error.

anehx:
@fkm Do you think this is still necessary?

fkm:
Good question. I stumbled upon this yesterday. Probably not.

anehx:
Closed since #376 refactored this bit

PR #355, opened 2019-06-20
fix(form): add missing validation function for form questions

anehx:


czosel:
@anehx How can I reproduce the issue that is fixed by this?

anehx:
`field.validate.perform()` on a FormQuestion field

PR #399, opened 2019-07-02
fix(ie11): fix checkboxes in IE11

anehx:


czosel:
Build seems to be failing ...

anehx:
@czosel should be fixed

anehx:
Actually I was thinking: do we really want to support IE11 in `ember-caluma`?

czosel:
Since IE 11 is still officially supported, and caluma somewhat targets enterprise users, I think we have to support IE 11. Also, how would we have fixed this issue outside of the addon?

anehx:
Well but it is also targeted for modern web applications.. However I agree that we should leave it *for now*. I just wanted to get that thought out.

The consumer can always overwrite the component...

PR #400, opened 2019-07-02
fix(uk-modal): add DDAU handler

fkm:

PR #406, opened 2019-07-04
fix(dummy): change cf-form to cf-content

fkm:
Small change to get the dummy page working again.

### DIFF
--- a/tests/dummy/app/templates/index.hbs
+++ b/tests/dummy/app/templates/index.hbs
@@ -1,5 +1,5 @@
 {{docs-hero}}
 
 <div class="uk-padding">
-  {{cf-form documentId=model}}
+  {{cf-content documentId=model}}
 </div>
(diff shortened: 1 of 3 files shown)


tests/dummy/app/templates/index.hbs
```diff
@@ -1,5 +1,7 @@
 {{docs-hero}}
 
 <div class="uk-padding">
-  {{cf-form documentId=model}}
+  {{#cf-content documentId=model as |content|}}
+    {{content.form}}
```
anehx: I think we can just use the default template, the block form is not really needed:

`{{cf-content documentId=model}}`

fkm: Ah nice :sunglasses:

fkm:
I'll wait until the pipline for #400 has finished and I could merge that branch. I'll then rebase this one and merge it afterwards. This fix was just so that I could test the other one.

PR #421, opened 2019-07-09
fix(deps): update jexl

anehx:
This should serve as a reminder that we need to update JEXL as soon as https://github.com/TomFrost/Jexl/pull/54 is merged. This PR works, however we should not update to another git dependency IMHO.

/cc @czosel


anehx:
Since https://github.com/TomFrost/Jexl/pull/54 is somehow not being merged I fixed our issue with a helper in this project as a workaround

PR #427, opened 2019-07-10
fix(form): fix saving of table answers

anehx:


czosel:
I restarted the CI, let's see if that helps.

PR #449, opened 2019-07-22
fix(performance): improve performance of state calculations

### LINKED ISSUE #420: More sync, less async
https://github.com/TomFrost/Jexl/blob/master/CHANGELOG.md#v210

This feature would allow us to move from ec-tasks to normal computed properties :tada: 

/cc @czosel

anehx:
* Use jexl's `evalSync` instead of `eval`
* Remove events in favor of computed properties

Fixes #420


anehx:
Fixes #420

PR #451, opened 2019-07-22
fix(form): fix premature access to validators which are not yet fetched

anehx:
/cc @JohnL17


czosel:
@anehx wanna make a new release? :slightly_smiling_face:

PR #472, opened 2019-08-09
fix(field): don't delete non-existent answers

fkm:
Currently, saving an empty value in a field with no previous value errors.

### DIFF
--- a/addon/lib/field.js
+++ b/addon/lib/field.js
@@ -329,7 +329,7 @@ export default Base.extend(ObjectQueryManager, {
     let response;
 
     if (value === null || value.length === 0) {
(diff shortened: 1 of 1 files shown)


czosel:
Looks good to me, but I think @anehx should still have a looks as well.

addon/lib/field.js
```diff
@@ -329,6 +329,10 @@ export default Base.extend(ObjectQueryManager, {
     let response;
 
     if (value === null || value.length === 0) {
+      if (!this.answer.uuid) {
```
anehx:
```suggestion
      if (this.isNew) {
```

PR #514, opened 2019-09-03
feat(options): add (un)archive to form-builder

fkm:
This works but - IMO - needs more discussion.

Screenshot_20190903_091152

### DIFF
--- a/addon/components/cfb-form-editor/question/options.js
+++ b/addon/components/cfb-form-editor/question/options.js
@@ -29,11 +29,12 @@ export default RenderComponent.extend({
       "optionRows",
       get(value, "edges.length")
         ? get(value, "edges").map(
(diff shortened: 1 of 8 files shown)


anehx:
Please add a test for this

addon/components/cfb-form-editor/question/options.js
```diff
@@ -96,6 +98,12 @@ export default RenderComponent.extend({
       this._update();
     },
 
+    archiveRow(row) {
```
anehx: Since this toggles the `isArchived` property I'd prefer a method name like `toggleRowArchived` to be more explicit in what this actually does.

fkm: Done :-)

PR #515, opened 2019-09-03
Fix caching

fkm:
The network policy property name was incorrect.

### DIFF
--- a/addon/components/cf-content.js
+++ b/addon/components/cf-content.js
@@ -1,18 +1,18 @@
+import { getOwner } from "@ember/application";
 import Component from "@ember/component";
-import layout from "../templates/components/cf-content";
-import { inject as service } from "@ember/service";
-import { computed } from "@ember/object";
 import { reads } from "@ember/object/computed";
+import { assert } from "@ember/debug";
+import { computed } from "@ember/object";
+import { inject as service } from "@ember/service";
 import { queryManager } from "ember-apollo-client";
 import { task } from "ember-concurrency";
+
 import Document from "ember-caluma/lib/document";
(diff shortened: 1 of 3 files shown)


fkm:
Floating dependencies are failing as we are using a private method in our code.  
https://github.com/projectcaluma/ember-caluma/blob/master/addon/utils/jexl.js#L17

@czosel Is our implementation really the correct one? I wonder if it is not possible to use the library without falling back to using 'private' methods.

czosel:
I think @anehx wrote that bit...

anehx:
> Floating dependencies are failing as we are using a private method in our code.
> [/addon/utils/jexl.js@`master`#L17](https://github.com/projectcaluma/ember-caluma/blob/master/addon/utils/jexl.js#L17) [](/projectcaluma/ember-caluma/blob/HEAD@{2019-09-04T06:47:02Z}/addon/utils/jexl.js#L17)
> 
> @czosel Is our implementation really the correct one? I wonder if it is not possible to use the library without falling back to using 'private' methods.

At the time I wrote that bit of code there was no way to parse the AST without using private methods. Is there now?

anehx:
@fkm I think we should change some of the commit types:

`fix(dummy)` => `docs(dummy)` dummy changes are always `docs`
`fix(docker)` => `chore(docker)` this is just a dependency update
`fix(gitignore)` => `chore(gitignore)` this is the perfect example for a chore / cleanup

The way the messages are now, we'd have a lot of irrelevant noise in our changelog. We should avoid that...

fkm:
Ok. I'll change those first thing in the afternoon.

anehx:
@fkm please rebase, #518 resolved the build issue :smile:

PR #519, opened 2019-09-04
fix(cfb-options): options cannot be deleted from db

fkm:
This change only presents the possibility to delete options as long as they are not saved to the database.

PS: This PR will not work as-is. It needs the changes from #515

### DIFF
--- a/addon/templates/components/cfb-form-editor/question/options.hbs
+++ b/addon/templates/components/cfb-form-editor/question/options.hbs
@@ -8,7 +8,7 @@
       model=row
     as |f|}}
       <div uk-grid class="uk-grid-small uk-flex">
-        {{#if (gt optionRows.length 1)}}
+        {{#if (and row.isNew (gt optionRows.length 1))}}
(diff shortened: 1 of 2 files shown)


anehx:
@fkm please rebase, #518 resolved the build issue :smile:

addon/templates/components/cfb-form-editor/question/options.hbs
```diff
@@ -8,7 +8,7 @@
       model=row
     as |f|}}
       <div uk-grid class="uk-grid-small uk-flex">
-        {{#if (gt optionRows.length 1)}}
+        {{#if row.isNew}}
```
anehx: I think this should be `(and row.isNew (gt optionRows.length 1))` otherwise it would be possible to remove the first and only row..

fkm: Good point!

anehx:
Also, there should be a test for this

PR #547, opened 2019-09-19
feat(input/options): hide archived options

fkm:


czosel:
Is this fragment only used in the form rendering part?

anehx:
@czosel Yes it is. The form builder does not know the "entity" `field`

czosel:
@anehx shall we cut a new release?

PR #553, opened 2019-09-21
fix: change disabled to readonly where applicable

czosel:
IIRC, the reasoning for this one goes like this:

* it’s impossible to copy text out of a disabled field in Firefox, which is problematic in one of our projects
* readonly fields look like they can be edited, which is bad UX

-> this is changing disabled fields to readonly ones, but styles them like disabled fields.

### DIFF
--- a/addon/components/cf-field/input/float.js
+++ b/addon/components/cf-field/input/float.js
@@ -9,11 +9,11 @@ import Component from "@ember/component";
 export default Component.extend({
   tagName: "input",
   classNames: ["uk-input"],
-  classNameBindings: ["field.isInvalid:uk-form-danger"],
+  classNameBindings: ["field.isInvalid:uk-form-danger", "disabled:uk-disabled"],
   attributeBindings: [
     "type",
     "step",
-    "disabled",
+    "disabled:readonly",
     "field.pk:name",
     "field.answer.value:value",
     "field.question.floatMinValue:min",
--- a/addon/components/cf-field/input/integer.js
+++ b/addon/components/cf-field/input/integer.js
@@ -9,11 +9,11 @@ import Component from "@ember/component";
 export default Component.extend({
   tagName: "input",
   classNames: ["uk-input"],
-  classNameBindings: ["field.isInvalid:uk-form-danger"],
+  classNameBindings: ["field.isInvalid:uk-form-danger", "disabled:uk-disabled"],
   attributeBindings: [
     "type",
     "step",
-    "disabled",
+    "disabled:readonly",
     "field.pk:name",
     "field.answer.value:value",
     "field.question.integerMinValue:min",
--- a/addon/components/cf-field/input/text.js
+++ b/addon/components/cf-field/input/text.js
@@ -9,10 +9,10 @@ import Component from "@ember/component";
 export default Component.extend({
   tagName: "input",
   classNames: ["uk-input"],
-  classNameBindings: ["field.isInvalid:uk-form-danger"],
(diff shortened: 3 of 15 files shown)


fkm:
The big question I remember was whether or not to implement this as an additional `readonly` property in Caluma. 

* There's the technical problem to overcome: checkboxes and radiobuttons don't have readonly.
* And then there's the question if it makes sense from a conceptual standpoint to have two different types of states.

My position is that it only makes sense to have one disabled/readonly/inactive state (i.e. the property we use in Caluma) and then use the best fit on a technical level (disabled for checkboxes and radiobuttons and readonly for the rest).

czosel:
That sounds very reasonable to me! What do you say @anehx ?

addon/components/cf-field/input/float.js
```diff
@@ -9,11 +9,14 @@ import Component from "@ember/component";
 export default Component.extend({
   tagName: "input",
   classNames: ["uk-input"],
-  classNameBindings: ["field.isInvalid:uk-form-danger"],
+  classNameBindings: [
+    "field.isInvalid:uk-form-danger",
+    "disabled:disabled"
```
anehx: Where is the styles of `.disabled`? Also, isn't there a class `uk-disabled`?

fkm: Good catch. It's missing from this MR. I directly put it into the project code.

```scss
.uk-input.disabled,
.uk-textarea.disabled {
  background-color: $form-disabled-background;
  color: $form-disabled-color;
  @if(mixin-exists(hook-form-disabled)) {@include hook-form-disabled();}
}
```

That should be included in the PR. I'll check why I didn't name this `uk-disabled`. I guess there's no such class in UIkit as they - IIRC - style through the element's attribute. I'll keep you posted.

fkm:
> I agree, this sound reasonable. However, we should still discuss whether we need to implement a proper readonly state (no input boxes but plain text).

I really like the idea of adding different components for the disabled/readonly state where we present the content of the form without any form controls. But that would - IMO -be a bigger project.

addon/components/cf-field/input/float.js
fkm: Yeah. They style through the attribute.
https://github.com/uikit/uikit/blob/5e0b52085fae878b08fc38107b34fa1629f8b22c/src/scss/components/form.scss#L290

I'll create a PR.

fkm: https://github.com/uikit/uikit/pull/3990

czosel:
I agree - for now keeping the input fields is the most straight forward solution and it has the additional advantage that the UI looks almost the same as in the editable state, which makes it easy to navigate the readonly view when the editable view is already known to the user.

addon/components/cf-field/input/float.js
anehx: I'm not sure if this PR will be accepted in the near future. Would it be a good idea to put @fkm your snipped above in the `ember-caluma` codebase?

anehx: Bump :point_up:

fkm: I agree that this PR is not going to be merged (any time soon). I'm ok with putting the styles into our code base.

czosel: @fkm can you update the PR?

fkm: Ok

fkm: Done

fkm:
Tests are updated.

PR #565, opened 2019-09-30
fix(navigation): fix navigation to show base form questions

### LINKED ISSUE #460: Main form questions are not rendered
When rendering a form with `{{cf-content}}` the questions of the main form are not displayed, only the ones of subforms. For example:

```
main-form
  -question-1
  -question-2
  -question-3 (subform)
    -question-4
    -question-5
```

In this case the navigation shows `question-3` and `question-4` and `question-5` are displayed. The two questions `question-1` and `question-2` are not accessible.

Yelinz:
Fixes #460

### DIFF
--- a/addon/lib/navigation.js
+++ b/addon/lib/navigation.js
@@ -59,7 +59,12 @@ export const NavigationItem = Base.extend({
    * @property {String} label
    * @accessor
    */
-  label: reads("fieldset.field.question.label"),
+  label: computed("fieldset.{field.question.label,form.name}", function() {
+    return this.getWithDefault(
+      "fieldset.field.question.label",
+      this.get("fieldset.form.name")
+    );
+  }),
 
   /**
    * The slug of the items form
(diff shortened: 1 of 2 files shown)


anehx:
I think you misunderstood the bug.. This "fix" will now show **all** questions in the navigation! However the bug is that if you have a form with subforms, there is no way to see the rendered main form, only the subforms.

/cc @kaldras or was this the actual task? If so it needs some discussion..

Yelinz:
> Will this now cause to render a navigation even if there are no subforms? 

There is no navigation when there are no subforms

> How is it structured - are all subforms indented a bit more or is the main form on the same level? 

They are all on the same level

> Could you append a screenshot?

Screenshot from 2019-10-01 11-06-56
With subform

Screenshot from 2019-10-01 11-07-30
Without subform

Yelinz:
Now with a divider between main and sub form navigation points
Screenshot from 2019-10-01 12-30-13

PR #577, opened 2019-10-14
fix(table): remove row documents on delete

fkm:
Hopefully, this will help with the historical API.


addon/components/cf-field/input/table.js
```diff
@@ -62,6 +63,12 @@ export default Component.extend(ComponentQueryManager, {
 
     yield this.onSave(remainingDocuments);
 
+    // Remove orphaned document.
+    yield this.apollo.mutate({
+      mutation: removeDocumentMutation,
+      variables: { input: { document: document.uuid } }
+    });
```
anehx: I think we need to unload it from the caluma store as well

fkm: Good catch. Done.

fkm:
I hope it will magically set the minus sign on the table row :smiley:

PR #586, opened 2019-10-22
feat(form): handle old values of dynamic questions

anehx:
This is now ready for review, but not for merging since I want to discuss some things with the backend dudes first..

### DIFF
--- a/addon/lib/field.js
+++ b/addon/lib/field.js
@@ -22,6 +22,7 @@ import saveDocumentFileAnswerMutation from "ember-caluma/gql/mutations/save-docu
 import saveDocumentDateAnswerMutation from "ember-caluma/gql/mutations/save-document-date-answer";
 import saveDocumentTableAnswerMutation from "ember-caluma/gql/mutations/save-document-table-answer";
 import removeAnswerMutation from "ember-caluma/gql/mutations/remove-answer";
+import getDocumentUsedDynamicOptionsQuery from "ember-caluma/gql/queries/get-document-used-dynamic-options";
 import { getAST, getTransforms } from "ember-caluma/utils/jexl";
 
 const TYPE_MAP = {
@@ -108,6 +109,10 @@ export default Base.extend({
       Question.create(getOwner(this).ownerInjection(), {
         raw: this.raw.question
       });
+
+    if (question.isDynamic) {
+      question.loadDynamicOptions.perform();
+    }
 
     this.set("question", question);
   },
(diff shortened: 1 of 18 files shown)


addon/templates/components/cf-field/input/radio.hbs
```diff
@@ -1,17 +1,19 @@
...
+      class="uk-radio uk-margin-small-right"
+      type="radio"
+      name={{field.pk}}
+      value={{option.slug}}
+      checked={{eq option.slug field.answer.value}}
+      disabled={{disabled}}
+      onchange={{action onSave option.slug}}
+    >
+    {{#if option.disabled}}
+      <strike class="uk-text-muted" title={{t "caluma.form.optionNotAvailable"}}>{{option.label}}</strike>
```
fkm: This element is obsolete.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strike

fkm: Perfekt!

addon/lib/field.js
```diff
@@ -207,6 +208,130 @@ export default Base.extend({
...
+      }));
+
+      const hasInvalidSelected = !selected.every(slug =>
+        options.find(option => option.slug === slug)
+      );
+
+      if (this.question.isDynamic && hasInvalidSelected) {
+        if (!this._fetchUsedDynamicOptions.lastSuccessful) {
+          // Fetch used dynamic options if not done yet
+          this._fetchUsedDynamicOptions.perform();
```
fkm: Shouldn't we wait here? (await or yield)

addon/lib/field.js
```diff
@@ -207,6 +208,130 @@ export default Base.extend({
...
+   * questions. It can either return null if no value is selected yet, an
+   * object for single choices or an array of objects for multiple choices.
+   *
+   * @property {Null|Object|Object[]} selected
+   * @accessor
+   */
+  selected: computed(
+    "value",
+    "options.@each.slug",
+    "usedDynamicOptions.@each.slug",
```
fkm: Shouldn't this be `options.[]` and `usedDynamicOptions.[]`?

If I understand the rest of the code correctly, the slugs themselves don't change.

addon/lib/field.js
anehx: This is a computed property, we can't await here

addon/lib/field.js
anehx: The only property of those options we need is `slug` so it makes no sense to observe the whole array

anehx: However, `"usedDynamicOptions.@each.slug"` is now obsolete, will remove

addon/lib/field.js
fkm: But don't we risk that `this.usedDynamicOptions` is still fetching when we enter the return statement a few lines below?

addon/lib/field.js
fkm: Ok

addon/lib/field.js
anehx: What would be the risk of that?

fkm: Probably, unexpected behavior. But I don't know. That's why I ask.

anehx: It's actually intended behaviour which will return the options without any used options. Since it will be recomputed "correctly" shortly after, this should not cause any problems.

fkm: Ok

PR #614, opened 2019-11-04
feat(license): switch license from MIT to LGPL-3.0-or-later

### LINKED ISSUE #613: License Change: LGPL-3.0-or-later
Following the license change of Caluma, we want to change the license of the Ember-Caluma, too.

## Goals
The same goals we evaluated in Caluma apply to the frontend:

- diverse uses and integration of Caluma

- Caluma to remain public as Free Software

- an active community

- to guarantee mutual benefit to our contributors

We found that in the frontend, we can and want to be more permissive for the following reasons:

- Since the frontend only works with the backend and the backend is GPL, the frontend can be LGPL without compromising our goals. There is no incentive to withhold changes in public interest.

- LGPL allows tight integration with existing frontends. We want to integrate the frontend directly in existing UIs that might not be compatible with GPL.

## What LGPL-3.0-or-later means for Ember-Caluma

The following summary is not legally binding, please refer to the [LGPL-3.0-or-later](https://spdx.org/licenses/LGPL-3.0-or-later.html):

You are free to modify and extend the Ember-Caluma as you wish. You can integrate the Ember-Caluma with any software you wish. If you publish or sell

- a modified or extended version of the frontend

- or an extension to frontend

you have to make the source-code of these modifications available under [LGPL-3.0-or-later](https://spdx.org/licenses/LGPL-3.0-or-later.html) or a compatible license.

## Comparison backend / frontend

- The GPL license of the backend does not affect integrations because it is server-side

- The LGPL license of the frontend does not affect integrations because LGPL is permissive

anehx:
BREAKING CHANGE: Since the LGPL license is more restrictive than MIT,
this is marked as breaking

Resolves #613

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,165 @@
-MIT License
+                   GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
 
-Copyright (c) 2018 Caluma
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
 
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
-furnished to do so, subject to the following conditions:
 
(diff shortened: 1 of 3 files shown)


ganwell:
@anehx I'll make a PR for the reuse-tool after this has been merged.

ganwell:
@anehx I have approved an only noticed afterwards:

chore(license) did not work in the backend, I had to change the commit to feat(license), because chores are ignored and BREAKING CHANGE is not read/respected.

anehx:
@ganwell fixed the commit message

PR #616, opened 2019-11-04
chore(reuse): define licenses of all files

ganwell:
* There might be files that have external licenses that I have missed. I used the heuristic: If the file has a commit log, that looks like development, it's not file form a external project

* I haven't added a linter, just created the dep5 and hoping for the best


ganwell:
I close this one. It is not important enough. Sorry for the noise.

PR #650, opened 2019-11-20
feat(form): allow empty values for answers

anehx:
This adds compatibility for
https://github.com/projectcaluma/caluma/pull/791. Instead of removing
answers with an empty value, we now save the empty value.

BREAKING CHANGE: Only works with caluma v4.0.0 (https://github.com/projectcaluma/caluma/pull/791)


anehx:
@czosel Hell yeah, especially those lines.. I wanted them gone for a loooong time :wink:

PR #654, opened 2019-11-21
chore(ci): add firefox to test matrix and use more compact reporter

anehx:
Running tests now looks like this:

Selection_019

Failed tests would be below the dot matrix with a clear error

### DIFF
--- a/testem.js
+++ b/testem.js
@@ -1,7 +1,10 @@
 module.exports = {
   test_page: "tests/index.html?hidepassed",
   disable_watching: true,
-  launch_in_ci: ["Chrome"],
+  parallel: "2",
+  reporter: "dot",
+  framework: "qunit",
+  launch_in_ci: ["Chrome", "Firefox"],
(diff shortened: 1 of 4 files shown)


testem.js
```diff
@@ -16,6 +19,12 @@ module.exports = {
         "--remote-debugging-port=0",
         "--window-size=1440,900"
       ].filter(Boolean)
+    },
+    Firefox: {
+      ci: [
+        "--headless",
```
fkm: Did you test this / does this work?

From the documentation the headless flag has only one dash:
https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Headless_mode

anehx: Interesting, yes I tested it and yes it works. However, you're right I'll change it to match the docs

anehx: Done

fkm: Good to know! :-)

PR #668, opened 2019-11-28
fix(cli): prepare apollo-client for next version

fkm:
> DEPRECATION: [ember-apollo-client] Deprecation:
>         The configuration option keepGraphqlFileExtension was not defined.
>         The current default is 'false', but it will change to 'true' after the next major release.

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -8,9 +8,8 @@ module.exports = function(defaults) {
   const app = new EmberAddon(defaults, {
     sassOptions: { implementation: sass },
     snippetPaths: ["tests/dummy/app/snippets"],
-    babel: {
-      plugins: ["@babel/plugin-proposal-object-rest-spread"]
-    }
+    babel: { plugins: ["@babel/plugin-proposal-object-rest-spread"] },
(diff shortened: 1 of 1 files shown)


anehx:
Needs discussion: Do we maybe want to set this to `true` since it's a) the new default and b) more verbose?

/cc @fkm @czosel @kaldras

fkm:
I'm for keeping it on `false`.

czosel:
I'd prefer `true` because that's what any user of the library will expect in the future.

anehx:
I'd also prefer `true` since this is the default and way more verbose

czosel:
@fkm why close?

fkm:
There was a 2:1 consensus.

czosel:
Do you think it's a lot of effort to make the suggested change?

fkm:
Adding the file extensions can - IMO - be done through a find-replace. We just have to do it before we upgrade to the upcoming version that changes the default setting.

czosel:
:+1: Wanna do it? :wink:

fkm:
I can do it sometime next week.

PR #720, opened 2019-12-24
Various form validation fixes

anehx:


addon/lib/answer.js
```diff
@@ -40,6 +41,10 @@ export default Base.extend({
     return this.raw.id ? decodeId(this.raw.id) : null;
   }),
 
+  isNew: computed("uuid", "value", function() {
+    return !this.uuid || isEmpty(this.value);
+  }),
```
fkm: Hmmm... What does `isNew` mean? 

Should an answer really be considered new if I remove the value?

anehx: New means that there is no value (yet)

anehx: I didn't want to change the API for something that still works the same way for the user but is only a small technical difference. Maybe `isEmpty` would be more clear. However, `isNew` works the same way as before since "emptying" the answer caused the removal of said question, now we just save an empty question.

fkm: Fair enough :ok_hand:

PR #723, opened 2019-12-27
fix(form): add proper table validation

anehx:


anehx:
This is pretty urgent, since no one but me works today I'll just trust myself and merge this :man_shrugging:

PR #788, opened 2020-02-20
feat(utils): increase slugify max-length to 150

fkm:
Match new max-length of backend.

https://github.com/projectcaluma/caluma/pull/942

### DIFF
--- a/addon/utils/slugify.js
+++ b/addon/utils/slugify.js
@@ -3,5 +3,5 @@ import slugify from "slugify";
 export default function(value) {
   return slugify(value.toLowerCase(), {
     remove: /((?![a-z0-9-\s]).)/g
-  }).substr(0, 50);
+  }).substr(0, 150);
 }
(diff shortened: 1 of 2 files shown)


addon/utils/slugify.js
```diff
@@ -3,5 +3,5 @@ import slugify from "slugify";
 export default function(value) {
   return slugify(value.toLowerCase(), {
     remove: /((?![a-z0-9-\s]).)/g
-  }).substr(0, 50);
+  }).substr(0, 150);
```
anehx: Why 150 instead of 255?

fkm: Because the backend hast 150.

anehx: Yes obviously, but why? I wasn't assigned in the backend PR so I'm asking now.

I'm just a bit concerned that we'll have to do this again as soon as 150 is not enough. So since the DB would allow 255 as a maximum I was wondering if it would be better to use the absolute maximum...

fkm: This is not the right place for this discussion. If you prefer a longer max-length, open a PR on the backend.

PR #792, opened 2020-02-27
feat(a11y): various improvements for accessibility

anehx:
* Add `for` property on labels
* Add hidden text and labels on icon buttons
* Add role heading and proper level on active navigation items

### DIFF
--- a/addon/templates/components/cf-navigation-item.hbs
+++ b/addon/templates/components/cf-navigation-item.hbs
@@ -1,13 +1,26 @@
-{{#link-to (query-params displayedForm=item.slug) class="uk-width-1-1"}}
-  <span title={{item.label}} class="uk-width-expand uk-text-truncate">{{item.label}}</span>
-  <span class="cf-navigation__item__icon cf-navigation__item__icon--{{item.state}}"></span>
-{{/link-to}}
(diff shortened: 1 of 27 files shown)


addon/templates/components/cf-navigation-item.hbs
```diff
@@ -1,13 +1,21 @@
-{{#link-to (query-params displayedForm=item.slug) class="uk-width-1-1"}}
-  <span title={{item.label}} class="uk-width-expand uk-text-truncate">{{item.label}}</span>
+<LinkTo @query={{hash displayedForm=item.slug}}>
+  {{#if this.active}}
+    <span role="heading" aria-level={{@level}} title={{item.label}} class="uk-width-expand uk-text-truncate">
```
fkm: I'm not quite sure if heading is the right role here.

Usually, H1 is the title of the current document. I understand the idea to include where the current page is in the complete form structure. Just not sure if this is ok.

You decide :smiley:

anehx: As discussed: We'll add an option to toggle whether to use the navigation as a heading or not since it may not always be wanted to be used as headings

PR #793, opened 2020-02-27
chore(deps): update dependencies

anehx:


addon/components/cfb-form-editor/general.js
```diff
@@ -5,7 +5,7 @@ import layout from "../../templates/components/cfb-form-editor/general";
 import { task, timeout } from "ember-concurrency";
 import { queryManager } from "ember-apollo-client";
 import validations from "../../validations/form";
-import v4 from "uuid/v4";
+import { v4 } from "uuid";
```
fkm: When Ember will be implementing Webpack in the future this syntax will prevent tree-shaking for this module.

The previous import syntax would allow Webpack to only include v4 in the bundle. With the new syntax all the version that are imported in the uuid index file will have to be included in the bundle.

But as this is still far off in the future and the syntax change probably enforced by a linter you can leave this as-is.

anehx: As discussed in person: the docs of uuid state that the former import is deprecated.

PR #805, opened 2020-03-11
feat(jexl): add info object to context

fkm:
Depends on https://github.com/projectcaluma/caluma/pull/978

```js
{
  form: rootFormSlug,
  info: {
    form: currentFormSlug,
    parent: { form: parentFormSlug },
    root: { form: rootFormSlug }
  }
}
```

`info.root.form == "root-form"`
`info.parent.form == "parent-form"`
`info.form == "current-form"`

Legacy JEXL expressions will still work through:
`form == "root-form"`

### DIFF
--- a/addon/lib/document.js
+++ b/addon/lib/document.js
@@ -139,7 +139,7 @@ export default Base.extend({
 
       return {
         form: this.rootForm.slug,
-        info: { root: { form: this.rootForm.slug } }
+        info: { root: { form: this.rootForm.slug } },
       };
     }
   ),
--- a/addon/lib/document.js
+++ b/addon/lib/document.js
@@ -137,7 +137,10 @@ export default Base.extend({
     function () {
       if (this.parentDocument) return this.parentDocument.jexlContext;
 
-      return { form: this.rootForm.slug };
+      return {
+        form: this.rootForm.slug,
+        info: { root: { form: this.rootForm.slug } }
+      };
     }
   ),
 
--- a/addon/lib/field.js
+++ b/addon/lib/field.js
@@ -9,6 +9,7 @@ import { task } from "ember-concurrency";
 import { all, resolve } from "rsvp";
 import { validate } from "ember-validators";
 import { queryManager } from "ember-apollo-client";
+import cloneDeep from "lodash.clonedeep";
 
 import { decodeId } from "ember-caluma/helpers/decode-id";
 
@@ -331,6 +332,31 @@ export default Base.extend({
   }),
 
   /**
+   * The field's JEXL context.
+   *
+   * Properties:
+   * - `form`: Legacy property pointing to the root form.
+   * - `info.form`: The form this question is attached to.
+   * - `info.parent.form`: The parent form if applicable.
+   * - `info.root.form`: The new property for the root form.
+   */
+  jexlContext: computed(
+    "document.jexlContext.{form,info.root.form}",
+    "fieldset.{form.slug,field.fieldset.form.slug}",
+    "question.{isHidden,isRequired}",
+    function() {
+      const context = cloneDeep(this.document.jexlContext);
+      context.info.form = this.get("fieldset.form.slug");
+
+      context.info.parent = this.get("fieldset.field.fieldset.form.slug")
+        ? { form: this.get("fieldset.field.fieldset.form.slug") }
+        : null;
+
+      return context;
+    }
+  ),
+
+  /**
    * Fields that are referenced in the `isHidden` JEXL expression
    *
    * If the value or hidden state of any of these fields change, the JEXL
@@ -400,10 +426,7 @@ export default Base.extend({
         getWithDefault(this, "fieldset.field.hidden", false) ||
         (this.hiddenDependencies.length &&
           this.hiddenDependencies.every(fieldIsHidden)) ||
-        this.document.jexl.evalSync(
-          this.question.isHidden,
-          this.document.jexlContext
-        )
+        this.document.jexl.evalSync(this.question.isHidden, this.jexlContext)
       );
     }
   ),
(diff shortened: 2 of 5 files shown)


addon/lib/field.js
```diff
@@ -400,10 +400,18 @@ export default Base.extend({
         getWithDefault(this, "fieldset.field.hidden", false) ||
         (this.hiddenDependencies.length &&
           this.hiddenDependencies.every(fieldIsHidden)) ||
-        this.document.jexl.evalSync(
-          this.question.isHidden,
-          this.document.jexlContext
-        )
+        this.document.jexl.evalSync(this.question.isHidden, {
+          ...this.document.jexlContext,
+          info: {
```
kaldras: As discussed, the `info` context also needs to be added to the optional property (line 429)

tests/unit/lib/document-test.js
```diff
@@ -126,7 +126,10 @@ module("Unit | Library | document", function(hooks) {
   test("computes the correct jexl context", async function(assert) {
     assert.expect(1);
 
-    assert.deepEqual(this.document.jexlContext, { form: "form" });
+    assert.deepEqual(this.document.jexlContext, {
+      form: "form",
+      info: { root: { form: "form" } }
```
kaldras: Would be nice if we have tests for `info.form` and `info.parent.form`

fkm: I'll look into it! :smiley:

addon/lib/field.js
```diff
@@ -330,6 +331,37 @@ export default Base.extend({
...
+   * - `info.root.form`: The new property for the root form.
+   */
+  jexlContext: computed(
+    "document.jexlContext.{form,info.root.form}",
+    "fieldset.{form.slug,field.fieldset.form.slug}",
+    function() {
+      const context = copy(this.document.jexlContext);
+
+      // This condition is needed to allow for partial states during testing.
+      if (this.get("fieldset")) {
```
anehx:
```suggestion
      if (this.fieldset) {
```

addon/lib/field.js
```diff
@@ -330,6 +331,37 @@ export default Base.extend({
...
+        context.info.form = this.get("fieldset.form.slug");
+      }
+
+      // Only attach the parent context if in a subform.
+      if (this.get("fieldset.form.slug")) {
```
anehx: This doesn't really make sense.. The field will always have a fieldset which also will always have a form. This condition will never be false.

I think you want to check whether the fieldset has a `field` property that would mean we are in a subform (the fieldset field is the form question)

addon/lib/field.js
anehx: Also, if this is only necessary for testing then maybe the tests should be adjusted to match reality

fkm: I'll look into adjusting the tests.

addon/lib/field.js
fkm: Good catch! Yeah. Somewhere along the line I must have lost the `fieldset.field` in the first getter.

tests/unit/lib/document-test.js
fkm: I added a test for `info.form`. I omitted `info.parent.form` for now as the existing dummy data doesn't include subforms. I'm not sure on how to best extend the dummy data without impacting the other tests that rely on this structure.

(Tests are in unit/lib/field.)

addon/lib/field.js
fkm: Should be better now. Testing both times for `fieldset.field.fieldset.form.slug`.

addon/lib/field.js
fkm: Removed the condition.

addon/lib/field.js
fkm: Added the computed property `jexlContext` which is used in `hidden` and `optional`.

addon/lib/field.js
```diff
@@ -330,6 +331,33 @@ export default Base.extend({
...
+   * - `info.root.form`: The new property for the root form.
+   */
+  jexlContext: computed(
+    "document.jexlContext.{form,info.root.form}",
+    "fieldset.{form.slug,field.fieldset.form.slug}",
+    function() {
+      const context = cloneDeep(this.document.jexlContext);
+      context.info.form = this.get("fieldset.form.slug");
+
+      // Only attach the parent context if in a subform.
```
anehx: Is this handled the same way in the backend or is the property just `None` if there is no parent?

fkm: From how I interpreted what I've read so far I thought that parent is missing completely if not in a subform. But I will check.

fkm: You were right! `parent` is implemented as a getter and thus always present.

https://github.com/projectcaluma/caluma/blob/master/caluma/caluma_form/structure.py#L28
Thank you and to @winged !

addon/lib/document.js
```diff
@@ -137,7 +137,10 @@ export default Base.extend({
     function() {
       if (this.parentDocument) return this.parentDocument.jexlContext;
 
-      return { form: this.rootForm.slug };
+      return {
+        form: this.rootForm.slug,
+        info: { root: { form: this.rootForm.slug } }
```
winged: We also have `info.parent.form`, which should also be implemented

addon/lib/field.js
winged: @anehx  JEXL (at least pyjexl) evaluates non-existing identifiers to `None` and crashes if you do anything "incompatible" with them (like continuing the dot notation, or adding transforms...). I'd add a `parent` here and just set it to `null` or `undefined` for consistency

addon/lib/document.js
anehx: This is implemented in the field since it's not possible to get that context per document

addon/lib/field.js
anehx: > @anehx JEXL (at least pyjexl) evaluates non-existing identifiers to `None` and crashes if you do anything "incompatible" with them (like continuing the dot notation, or adding transforms...). I'd add a `parent` here and just set it to `null` or `undefined` for consistency

@fkm Alright, let's do that then

fkm:
# Do not delete branch after merge!

addon/lib/field.js
fkm: Done

anehx:
@fkm could you rebase this again? @kaldras can you take a look at the changes requested by you?

PR #893, opened 2020-06-03
fix(i18n): only use primary locale in http headers

anehx:


czosel:
Out of curiosity: What's the difference? (Which language was sent before?)

anehx:
> Out of curiosity: What's the difference? (Which language was sent before?)

`intl.locale` always returns an array whereas `intl.primaryLocale` returns the first:

https://ember-intl.github.io/ember-intl/docs/guide/service-api#locale
https://ember-intl.github.io/ember-intl/docs/guide/service-api#primarylocale-readonly

PR #1003, opened 2020-08-26
chore(deps): update various dependencies and resolve deprecations

anehx:
Closes #999 
Closes #993 
Closes #990 
Closes #978 
Closes #968

PR #1004, opened 2020-08-26
chore(ci): add dependency linting

anehx:
Depends on #1003 

This tool makes sure that only one version of a dependency is installed. It can also be used to say what version. This will help us maintain this addon since the dependency problem already bit us in the ass a couple of times..

### DIFF
--- a/config/dependency-lint.js
+++ b/config/dependency-lint.js
@@ -1,0 +1,13 @@
+"use strict";
+
+module.exports = {
+  generateTests: false,
+  allowedVersions: {
+    // https://github.com/jasonmit/ember-cli-moment-shim/issues/183
+    moment: "2.24.0",
+
+    // EC v3+ heavily relies on Proxy which is not supported by IE11
+    "ember-changeset": "^2.2.4",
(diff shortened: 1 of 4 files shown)


config/dependency-lint.js
```diff
@@ -0,0 +1,13 @@
...
+
+module.exports = {
+  generateTests: false,
+  allowedVersions: {
+    // https://github.com/jasonmit/ember-cli-moment-shim/issues/183
+    moment: "2.24.0",
+
+    // EC v3+ heavily relies on Proxy which is not supported by IE11
+    "ember-changeset": "^2.2.4",
+    "ember-changeset-validations": "^2.2.1",
```
czosel, line 11: Should we also list the GraphQL dependency mentioned in the dependabot config here?

czosel (approved):
Awesome, I didn't know about dependency-lint! :100: Just one question.

config/dependency-lint.js
anehx: Not sure. Those dependencies in here should not be updated. However, `graphql-tools` should be updated but I didn't have the time to do it right now..

PR #1007, opened 2020-08-27
refactor(slug): use slugify options instead of manual operations

anehx:


JohnL17 (approved):

czosel (approved):
:+1: 

One other thing which is somewhat related that I'd like to improve: Could we replace `ä`, `ö`, `ü` by `ae`, `oe`, `ue`?

anehx:
I think we'd need to pass the locale for that to work: https://github.com/simov/slugify/blob/master/config/locales.json

czosel:
Ah, I didn't know there is locale support. Would be awesome to have this!

anehx:
Done!

czosel:
:100:

PR #1016, opened 2020-09-01
fix(embed): render power select in place

czosel:
Without this no options are rendered if ember-caluma is embedded.


czosel:
Using `renderInPlace` seems to have some drawbacks when it comes to positioning flexibility - but I have the feeling we don't need this flexibility, so changing this globally should be fine.

anehx (approved):

PR #1048, opened 2020-09-29
Cases query

velrest:


addon/caluma-query/queries/base.js
```diff
@@ -18,12 +18,14 @@ export default class BaseQuery {
     pageSize = null,
     processNew = async (items) => items,
     processAll = async (items) => items,
+    queryOptions = {},
```
anehx, line 21: This needs to be documented

addon/caluma-query/queries/index.js
```diff
@@ -1,7 +1,12 @@
 import WorkItemQuery from "./work-item";
+import CaseQuery from "./case";
 
 export function allWorkItems(options) {
   return new WorkItemQuery(options);
 }
 
-export default { allWorkItems };
+export function allCases(options) {
```
anehx, line 8: Document it...

anehx (changes requested):
Code looks good but documentation is inadequate, please fix that.

addon/caluma-query/queries/base.js
velrest: Done :+1:

addon/caluma-query/queries/index.js
velrest: Also done

anehx:
Now we still have failing tests

velrest:
@anehx should be fixed now

anehx (approved):

PR #1054, opened 2020-10-05
feat(cfb): add seperation between archived and active forms

Yelinz:
image

Add separate list for archived forms and labels for the form state

### DIFF
--- a/addon/components/cfb-form-list/item.js
+++ b/addon/components/cfb-form-list/item.js
@@ -1,10 +1,33 @@
 import Component from "@ember/component";
 import layout from "../../templates/components/cfb-form-list/item";
+import { inject as service } from "@ember/service";
 
-export default Component.extend({
-  layout,
+export default class componentsCfbFormItemListItem extends Component {
+  @service intl;
 
-  tagName: "li",
+  layout = layout;
 
-  classNames: ["uk-text-nowrap", "uk-flex", "uk-flex-middle"],
-});
+  tagName = "li";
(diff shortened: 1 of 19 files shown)


anehx:
Hm maybe tabs (or a button group filter) would be better? I mean what's the use case for wanting to see all forms at once?

Yelinz:
> Hm maybe tabs (or a button group filter) would be better? I mean what's the use case for wanting to see all forms at once?

image

I dont see the use case aswell and it looks nicer now

Yelinz:
image
This is how the filters look like

addon/components/cfb-form-list/item.js
```diff
@@ -1,10 +1,33 @@
...
-  layout,
+export default class componentsCfbFormItemListItem extends Component {
+  @service intl;
 
-  tagName: "li",
+  layout = layout;
 
-  classNames: ["uk-text-nowrap", "uk-flex", "uk-flex-middle"],
-});
+  tagName = "li";
```
anehx, line 10: Use a glimmer component so this won't be necessary, this component is easy to migrate..

anehx (changes requested):

addon/components/cfb-form-list/item.js
Yelinz: Changed

czosel:
Tests are failing ...

anehx (approved):

PR #1055, opened 2020-10-05
fix(deps): update jexl to 2.3.0

anehx:
https://github.com/TomFrost/Jexl/commit/eecff1632c374ec45385ba606761c603d6f53222 this commit broke our jexl utility


czosel (approved):
:100:

czosel:
Is the new "functions" feature in jexl something we should take a closer look at? Are they nicer than transforms?

PR #1065, opened 2020-10-07
Fix test suite

Yelinz:


velrest (approved):

Yelinz:
Seems the failure is caused by ember-gestures, which is used by ember-toggle, which comes from ember-uikit. Their teardown function doesnt seem to be compatible with ember 3.22

czosel:
Can you open a PR in ember-gestures that fixes this @Yelinz ? cc @anehx

Yelinz:
> Can you open a PR in ember-gestures that fixes this 

https://github.com/emberjs/ember.js/issues/19192

Ive narrowed it down to a problem in ember, it should be a deprecation but it throws an error

Yelinz:
Waiting for a new release in ember then the pipelines will be fixed

czosel:
Should we mark the ember-release stage as optional in order to move forward?

Yelinz:
> Should we mark the ember-release stage as optional in order to move forward?

Imo yes, the test should run but we say latest lts support which is 3.20

anehx:
Temporarily yes, but we always need to support the latest version. Support for latest LTS means back to that version not only that version..

velrest (approved):

PR #1066, opened 2020-10-07
fix(form): don't throw when typename is undefined in cf-field/input

czosel:
In one of our projects we ran into the issue that `typename` was
undefined during the loading of the page. This fix makes sure that we
don't fail in this situation.

This also includes a drive-by fix that prevents incorrect component
registration warnings.


anehx (approved):

anehx:
Tests are failing

PR #1069, opened 2020-10-14
refactor(caluma-query): allow `queryOptions` to be set on `fetch`

velrest:
This allows dynamic filtering via `queryOptions` since the decorator has no way to recompute the value of queryOptions.
Im still leaving in the decorator param since this can be still usefull for options that do not change.


velrest:
Pipeline will fail for now because of the issues mention in #1065

Yelinz (approved):

anehx (approved):

PR #1085, opened 2020-10-23
chore: pin ember-truth-helpers

Yelinz:
Pin truth helpers to fix the pipeline


anehx:
What exactly is the issue?

Yelinz:
Ember truth helpers were updated to 3.0.0, some packages updated but ember-validated-form still has the older version

anehx:
If it doesn't create any errors then configure dependency lint: https://github.com/salsify/ember-cli-dependency-lint#allowed-versions

Yelinz:
Just configuring dependency lint doesnt seem to work

Yelinz:
The pinning can be removed after ember-validated-forms has updated their version of ember-truth-helpers

anehx:
> The pinning can be removed after ember-validated-forms has updated their version of ember-truth-helpers

You're aware that we are the maintainers of that addon, right? :wink:

Yelinz:
Yes but that addon needs some maintenance done

anehx:
Ah come on, that update will cost you about 10 minutes.. I just checked the changelog of `ember-truth-helpers` there are no breaking changes other than node support drops and an update to babel 7, which is already in use.

Yelinz:
Updated it but now the problem lies with ember-cli-addon-docs which doesnt have a newer version

package.json
```diff
@@ -65,20 +65,20 @@
...
-    "@adfinis-sygroup/semantic-release-config": "2.1.4",
+    "@adfinis-sygroup/semantic-release-config": "3.0.1",
     "@babel/plugin-proposal-object-rest-spread": "7.12.1",
     "@ember/jquery": "1.1.0",
     "@ember/optional-features": "2.0.0",
     "@glimmer/component": "1.0.2",
     "babel-eslint": "10.1.0",
     "broccoli-asset-rev": "3.0.0",
-    "ember-cli": "3.22.0",
+    "ember-cli": "3.21.2",
```
anehx, line 75: Why downgrade?

Yelinz: So its equal with ember-source and the ember cli update command works for the next update

anehx: Okay, next time open a separate PR for this. This doesn't have anything to do with pinning ember-truth-helpers

PR #1098, opened 2020-10-29
feat(lib): add a debug transform which logs to console

### LINKED ISSUE #1095: Implement debug transform
See projectcaluma/caluma/issues/1248 (and backend implementation in projectcaluma/caluma/issues/1249)

> During form configuration, it may sometimes be difficult to decide if an expression is correct or not, or if a value contains the expected data.
> For this reason, I propose introducing a `debug` transform that will log the transformed value (`console.log()` in JS, `log.info()` in python), then return the value without modification.
> This would allow you to write expressions like `'foo'|answer|debug == 'foo-selection'` and see if the expression left of the transform actually does what you think. It will also allow some exploratory experimenting for people who are not familiar with available variables.

fkm-adfinis:
The passed value is logged to console and then passed along
for further chaining.

Closes #1095

### DIFF
--- a/addon/lib/document.js
+++ b/addon/lib/document.js
@@ -121,6 +121,10 @@ export default Base.extend({
       return Array.isArray(arr) ? arr.map((obj) => obj[key]) : null;
     });
     documentJexl.addBinaryOp("intersects", 20, intersects);
+    documentJexl.addTransform("debug", (any, label = "JEXL Debug") => {
+      console.debug(`${label}:`, any);
+      return any;
+    });
 
     return documentJexl;
   }),


winged:
Note: Converted to draft, as there might be an issue with the dependency graph story if used incorrectly:

`'foo'|answer|debug` would certainly work, but `foo|debug|answer` could cause issues, as the transform extractor would not recognize the lefthand side as a valid input, and will crash. This needs to be addressed as well (similar to the backend implementation)

addon/lib/document.js
```diff
@@ -121,6 +121,10 @@ export default Base.extend({
       return Array.isArray(arr) ? arr.map((obj) => obj[key]) : null;
     });
     documentJexl.addBinaryOp("intersects", 20, intersects);
+    documentJexl.addTransform("debug", (any) => {
+      console.log("Jexl Debug: ", any);
```
anehx, line 125: `console.debug` might be the better choice here

fkm-adfinis: Good idea!

anehx: A little bit of context (e.g the question slug) would be nice

fkm-adfinis: Is there a way to get at the question slug?

anehx: Just checked - sadly no. So ignore my comment ^^

anehx (approved):

addon/lib/document.js
```diff
@@ -121,6 +121,10 @@ export default Base.extend({
       return Array.isArray(arr) ? arr.map((obj) => obj[key]) : null;
     });
     documentJexl.addBinaryOp("intersects", 20, intersects);
+    documentJexl.addTransform("debug", (any) => {
```
anehx, line 124: What we could do is allowing an argument for debug that will be used to "label" the console debug: `|debug("test")` would result in `test: yourvalue`. As default we could use your proposed "JEXL Debug"

addon/lib/document.js
fkm-adfinis: Would have been nice though :smiley:

addon/lib/document.js
fkm-adfinis: Nice idea!

fkm-adfinis: Done

winged: Nice idea! Will replicate the same in the backend as well, then we can un-draft this and get it in master :+1:

PR #1116, opened 2020-11-17
Various chore tasks

anehx:


.dependabot/config.yml
```diff
@@ -18,20 +18,20 @@ update_configs:
...
       - match:
           dependency_name: "ember-changeset"
-          version_requirement: ">=3"
+          version_requirement: ">= 3"
       - match:
           dependency_name: "ember-changeset-validations"
-          version_requirement: ">=3"
+          version_requirement: ">= 3"
 
       # graphql-tools v5+ breaks generic mocking
```
czosel, line 26: this is not correct anymore, is it?

czosel (approved):
Otherwise LGTM :+1:

.dependabot/config.yml
anehx: Sure it is, we still have to support IE11 which sadly didn't magically develop the ability to use `Proxy` :wink:

czosel: I'm confused because the comment says "graphql-tools v5+ breaks generic mocking" and the version requirement was changed from "<=4" to ">= 5"

anehx: As discussed directly: the previous configuration was wrong since it's an **ignore** configuration. We want dependabot to ignore v5+

PR #1131, opened 2020-11-25
feat(nav): implement navigation states for default answers

anehx:
Navigation items which have fields that have only default answers will be greyed out to indicate that the user didn't do anything there but will still show the state (check, times or question mark) to indicate the state of the data in there:

Screenshot from 2020-11-25 14-47-23

### DIFF
--- a/addon/lib/field.js
+++ b/addon/lib/field.js
@@ -8,6 +8,7 @@ import { queryManager } from "ember-apollo-client";
 import { task } from "ember-concurrency";
 import { validate } from "ember-validators";
 import cloneDeep from "lodash.clonedeep";
+import isEqual from "lodash.isequal";
 import { all, resolve } from "rsvp";
 
 import saveDocumentDateAnswerMutation from "ember-caluma/gql/mutations/save-document-date-answer";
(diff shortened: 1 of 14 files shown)


addon/lib/field.js
```diff
@@ -187,6 +187,39 @@ export default Base.extend({
...
+    "question.{isTable,defaultValue}",
+    function () {
+      if (!this.value || !this.question.defaultValue) return false;
+
+      const value = this.question.isTable
+        ? this.value.map((doc) => doc.flatAnswerMap)
+        : this.value;
+
+      return (
+        JSON.stringify(value) === JSON.stringify(this.question.defaultValue)
```
fkm-adfinis, line 207: This worried me a bit as the order of object properties is not guaranteed. But from what I've learned they should respect the order they were set in.

So, my question becomes: Is the order of `this.fields` dependent on configuration or the order in which they were filled in by the user? If it is by configuration, this comparison should not be a problem.

anehx: Very good point. The order of the fields is based on configuration, however the order of the default answer is not.. I'll have to find another way of comparing those. Do you have experience with https://lodash.com/docs/4.17.15#isEqual ?

fkm-adfinis: I remember using it a lot with React to check if the properties had changed. *I don't remember having any issues with it.* But, that was a long time ago and maybe I suppressed a bad memory including isEqual :smile: 

But yeah. I agree that this sounds like a good idea!

anehx: Tested it and it works. I also added a test, will push soon. Good catch!

fkm-adfinis: :+1:

fkm-adfinis (approved):

kaldras:
Very nice change! I believe this will bring more clarity for the user to what work should or can be done and the overall state of the form! :+1:

kaldras (approved):

PR #1152, opened 2020-12-11
feat(cfb): add input for default values

fkm-adfinis:
This will create/update/delete default answers which will be used when rendering a form to fill out.

# Screenshots
## Text  
Text

Textarea

## Numbers
Integer

Float

## Date
Date

## Choice
Choice

MultipleChoice

## Table
Table

### DIFF
--- a/addon/components/cfb-form-editor/question.js
+++ b/addon/components/cfb-form-editor/question.js
@@ -4,6 +4,7 @@ import Component from "@ember/component";
 import { computed, get } from "@ember/object";
 import { reads } from "@ember/object/computed";
 import { inject as service } from "@ember/service";
+import { camelize } from "@ember/string";
 import { queryManager } from "ember-apollo-client";
 import Changeset from "ember-changeset";
 import lookupValidator from "ember-changeset-validations";
@@ -14,9 +15,16 @@ import { all } from "rsvp";
 import layout from "../../templates/components/cfb-form-editor/question";
 
 import addFormQuestionMutation from "ember-caluma/gql/mutations/add-form-question";
+import removeDefaultAnswerMutation from "ember-caluma/gql/mutations/remove-default-answer";
 import saveCalculatedFloatQuestionMutation from "ember-caluma/gql/mutations/save-calculated-float-question";
 import saveChoiceQuestionMutation from "ember-caluma/gql/mutations/save-choice-question";
 import saveDateQuestionMutation from "ember-caluma/gql/mutations/save-date-question";
+import saveDefaultDateAnswerMutation from "ember-caluma/gql/mutations/save-default-date-answer";
+import saveDefaultFloatAnswerMutation from "ember-caluma/gql/mutations/save-default-float-answer";
+import saveDefaultIntegerAnswerMutation from "ember-caluma/gql/mutations/save-default-integer-answer";
+import saveDefaultListAnswerMutation from "ember-caluma/gql/mutations/save-default-list-answer";
+import saveDefaultStringAnswerMutation from "ember-caluma/gql/mutations/save-default-string-answer";
+import saveDefaultTableAnswerMutation from "ember-caluma/gql/mutations/save-default-table-answer";
 import saveDynamicChoiceQuestionMutation from "ember-caluma/gql/mutations/save-dynamic-choice-question";
 import saveDynamicMultipleChoiceQuestionMutation from "ember-caluma/gql/mutations/save-dynamic-multiple-choice-question";
 import saveFileQuestionMutation from "ember-caluma/gql/mutations/save-file-question";
@@ -51,6 +59,15 @@ export const TYPES = {
   StaticQuestion: saveStaticQuestionMutation,
   DateQuestion: saveDateQuestionMutation,
   CalculatedFloatQuestion: saveCalculatedFloatQuestionMutation,
+};
+
+const TYPES_ANSWER = {
+  StringAnswer: saveDefaultStringAnswerMutation,
+  IntegerAnswer: saveDefaultIntegerAnswerMutation,
+  FloatAnswer: saveDefaultFloatAnswerMutation,
+  DateAnswer: saveDefaultDateAnswerMutation,
+  ListAnswer: saveDefaultListAnswerMutation,
+  TableAnswer: saveDefaultTableAnswerMutation,
 };
 
 export default Component.extend({
@@ -95,6 +112,7 @@ export default Component.extend({
             floatMaxValue: null,
             minLength: null,
             maxLength: null,
+            defaultAnswer: null,
             options: [],
             rowForm: {},
             subForm: {},
@@ -302,6 +320,39 @@ export default Component.extend({
     );
   }),
 
+  saveDefaultAnswer: task(function* (question, changeset) {
+    const answer = changeset.get("defaultAnswer");
+
+    if (!answer) {
+      return;
+    }
+
+    const valueKey = camelize(answer.__typename.replace(/Answer$/, "Value"));
+    const value = answer[valueKey];
+
+    // We need to map the UUIDs of the document's if the user didn't touch
+    // the default answer and thus never triggered the onUpdate action.
+    if (answer.__typename === "TableAnswer" && typeof value[0] !== "string") {
+      return;
+    }
+
+    const isAddMutation = value !== null && value.length !== 0;
+
+    // Save or remove depending on the value.
+    const mutation = !isAddMutation
+      ? removeDefaultAnswerMutation
+      : TYPES_ANSWER[answer.__typename];
+
+    const input = { question: question.slug };
+
+    // The remove mutation must not include a value.
+    if (isAddMutation) {
+      input.value = value;
+    }
+
+    yield this.apollo.mutate({ mutation, variables: { input } });
+  }),
+
   submit: task(function* (changeset) {
     try {
       yield this.saveOptions.perform(changeset);
@@ -315,6 +366,8 @@ export default Component.extend({
         },
         `save${typename}.question`
       );
+
+      yield this.saveDefaultAnswer.perform(question, changeset);
 
       if (!this.slug) {
         // This is a new question which must be added to the form after creating it
--- a/addon/components/cfb-form-editor/question/default.js
+++ b/addon/components/cfb-form-editor/question/default.js
@@ -1,0 +1,98 @@
+import { getOwner } from "@ember/application";
+import { action, computed, get } from "@ember/object";
+import { inject as service } from "@ember/service";
+import RenderComponent from "ember-validated-form/components/validated-input/-themes/uikit/render";
+
+import { TYPE_MAP } from "ember-caluma/lib/field";
+import layout from "ember-caluma/templates/components/cfb-form-editor/question/default";
+
+export default class CfbFormEditorQuestionDefault extends RenderComponent {
+  layout = layout;
+
+  @service router;
+
+  @computed("model.{id,__typename}")
+  get question() {
+    const raw = {
+      ...this.model.get("data"),
+      ...this.model.get("change"),
+      // While we want the real value, typename, etc. of the question,
+      // the question should never be required or hidden in the form-builder.
+      // We need to set a value here as no value will lead to a Jexl error.
+      isRequired: "false",
+      isHidden: "false",
+    };
+
+    if (
+      ["ChoiceQuestion", "MultipleChoiceQuestion"].includes(
+        this.model.__typename
+      )
+    ) {
+      // Use Power Select for choice questions to save space.
+      raw.meta = { widgetOverride: "cf-field/input/powerselect" };
+
+      const key = this.model.__typename
+        .replace(/^./, (match) => match.toLowerCase())
+        .replace("Question", "Options");
+
+      raw[key] = raw.options;
+      delete raw.options;
+    }
+
+    return raw;
+  }
+
+  get field() {
+    return getOwner(this)
+      .factoryFor("caluma-model:field")
+      .create({
+        raw: {
+          question: this.question,
+          // The value depends on where it comes from. If there is a default
+          // value present on load the `value.content` will be set. After an
+          // update through this component the value will be a POJO on `value`.
+          answer: (this.value && this.value.content) ||
+            this.value || {
+              id: btoa(`Answer:dv-answer-${this.model.slug}`),
+              __typename: TYPE_MAP[this.model.__typename],
+            },
+        },
+        // In order to mock the field, we need a document instance and its form.
+        document: getOwner(this)
+          .factoryFor("caluma-model:document")
+          .create({
+            raw: {
+              id: btoa(`Document:dv-document-${this.model.slug}`),
+              rootForm: {
+                slug:
+                  // There is no currentRoute in the integration tests.
+                  get(this, "router.currentRoute.attributes.formSlug") ||
+                  "dv-form",
+                __typename: "Form",
+              },
+              forms: [],
+              __typename: "Document",
+            },
+          }),
+      });
+  }
+
+  @action async onUpdate(value) {
+    this.set("field.answer.value", value);
+
+    await this.field.validate.perform();
+
+    this.update({
+      [this.field.answer._valueKey]: this.field.answer.serializedValue,
+      __typename: this.field.answer.__typename,
+      id: this.field.answer.id,
+    });
+
+    if (this.field.errors.length > 0) {
+      // The errors must be pushed after calling update.
+      this.model.pushErrors(this.name, ...this.field.errors);
+    }
+
+    this.setDirty();
+  }
+}
--- a/addon/gql/queries/form-editor-question.graphql
+++ b/addon/gql/queries/form-editor-question.graphql
@@ -1,4 +1,6 @@
 # import * from 'ember-caluma/gql/fragments/question-info'
+# import * from 'ember-caluma/gql/fragments/field-question'
+# import SimpleAnswer from 'ember-caluma/gql/fragments/field-answer'
 
 query FormEditorQuestion($slug: String!) {
   allQuestions(slug: $slug) {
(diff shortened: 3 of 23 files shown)


fkm-adfinis:
After working around the issue with focus-trap I found myself struggling with the test suite.

I get the following error but I fail to find what I forgot. I do get into `handleAnswer` but with an empty object as parameter. I stepped through the whole process and have to admit that I don't really understand what is happening under the hood in Apollo.

Screenshot_20201222_123034

Do you have an idea, what I could be missing?

addon/components/cfb-form-editor/question/default.js
```diff
@@ -0,0 +1,94 @@
...
+
+import { TYPE_MAP } from "ember-caluma/lib/field";
+
+export default RenderComponent.extend({
+  layout,
+
+  subComponent: computed("model.__typename", function () {
+    const type = this.model.__typename.replace("Question", "").toLowerCase();
+    return ["choice", "multiplechoice"].includes(type) ? "powerselect" : type;
+  }),
```
anehx, line 15: Why aren't we using the `cf-field` components?

addon/gql/queries/form-editor-question.graphql
```diff
@@ -1,4 +1,35 @@
 # import * from 'ember-caluma/gql/fragments/question-info'
+# import * from 'ember-caluma/gql/fragments/field-question'
+
+fragment SimpleAnswer on Answer {
```
anehx, line 4: We already have that exact fragment in `gql/fragments/simple-answer.graphql`, why duplicate it?

package.json
```diff
@@ -126,6 +126,7 @@
     "typeface-oxygen-mono": "1.1.13"
   },
   "resolutions": {
+    "**/focus-trap": "6.2.1",
```
anehx, line 129: Can we add a comment on why we need that? If we just keep adding resolutions without documented reasoning, they won't ever disappear..

anehx (changes requested):

addon/components/cfb-form-editor/question.js
```diff
@@ -290,6 +308,35 @@ export default Component.extend({
...
+
+    // Save or remove depending on the value.
+    const mutation =
+      value === null
+        ? removeDefaultAnswerMutation
+        : TYPES_ANSWER[answer.__typename];
+
+    const input = {
+      question: question.slug,
+      clientMutationId: v4(),
```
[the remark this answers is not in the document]
fkm-adfinis: Gladly :-)

addon/components/cfb-form-editor/question/default.js
```diff
@@ -0,0 +1,94 @@
+import { getOwner } from "@ember/application";
+import { computed } from "@ember/object";
+import RenderComponent from "ember-validated-form/components/validated-input/-themes/uikit/render";
+
+import layout from "../../../templates/components/cfb-form-editor/question/default";
+
+import { TYPE_MAP } from "ember-caluma/lib/field";
+
+export default RenderComponent.extend({
```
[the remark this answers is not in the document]
fkm-adfinis: I'll give it a shot :-) I'm a bit afraid because of the tracking. But you're right. It's a new component.

addon/components/cfb-form-editor/question/default.js
fkm-adfinis: I'll have to check this decision. This probably comes from the time before I mocked lib/field instead of just lib/answer and lib/question.

addon/gql/queries/form-editor-question.graphql
fkm-adfinis: Oversight. I'll use the fragment.

package.json
fkm-adfinis: Good point. I'll add a note to the read-me.

addon/components/cfb-form-editor/question/default.js
```diff
@@ -0,0 +1,94 @@
...
+    }
+  ),
+
+  field: computed(
+    "answer",
+    "model.__typename",
+    "question",
+    "value.content",
+    function () {
+      const document = getOwner(this)
```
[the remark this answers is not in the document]
fkm-adfinis: I'll add a comment.

addon/components/cfb-form-editor/question.js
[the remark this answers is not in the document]
fkm-adfinis: Done

addon/components/cfb-form-editor/question/default.js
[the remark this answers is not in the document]
fkm-adfinis: Done. Not sure if this is enough. I'll add more if you want.

addon/gql/queries/form-editor-question.graphql
fkm-adfinis: Done

package.json
fkm-adfinis: Done

addon/components/cfb-form-editor/question/default.js
fkm-adfinis: Using `cf-field` is too specific.

Screenshot_20201223_123038

As a compromise I switched to `{{component (get-widget` which seems to work. Not sure, if I should set a Power Select override for (Multiple)Choice. I'll look into how I would do that.

Screenshot_20201223_125110

Screenshot_20201223_125202

addon/components/cfb-form-editor/question/default.js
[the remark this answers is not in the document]
fkm-adfinis: First impression: Works mostly. I'll have to tweak a few things but it's not as bad as I feared.

addon/components/cfb-form-editor/question/default.js
fkm-adfinis: Ok. I found a simple way to set Power Select.

addon/components/cfb-form-editor/question/default.js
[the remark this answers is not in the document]
fkm-adfinis: I had to introduce a `@computed` decorator for it to work. But since adding it I didn't have any problems.

addon/components/cfb-form-editor/question/default.js
```diff
@@ -0,0 +1,83 @@
...
+
+  get field() {
+    // In order to mock the field, we need a document instance and its form.
+    const document = getOwner(this)
+      .factoryFor("caluma-model:document")
+      .create({
+        raw: {
+          id: btoa("Document:dv-document"),
+          rootForm: {
+            slug: "dv-form",
```
fkm-adfinis, line 42: I was wondering, if there is a good way to get the currently selected form.

There seems to be a way to get to the slug through private properties of the Ember components. But aside from the private properties that might break with every update it would also include a request to the API.

Do you have an idea, how we could access the form from withing `cfb-form-editor/question`?

README.md
```diff
@@ -67,6 +67,21 @@ yarn update-fragment-types
...
+  https://github.com/projectcaluma/ember-caluma/pull/1152  
+  Focus trap now throws an error if there is no tabbable element on initialization. This breaks the modal dialogs as they are hidden on initialization.
+* **ember-truth-helpers**  
+  https://github.com/projectcaluma/ember-caluma/pull/1085  
+  The update to 3.0.0 lead to a situation where we couldn't resolve conflicting sub-dependencies. (incl. ember-validated-fom)
+* **ember-get-config**  
+  https://github.com/projectcaluma/ember-caluma/pull/1065  
+  We pinned this package as there seems to have been a problem upstream with a thrown error instead of a deprecation warning.
+* **ember-test-waiters**  
+  https://github.com/projectcaluma/ember-caluma/commit/b0531664e95a529f21e44cdde9b9d782f417d00d
```
anehx, line 83: Awesome, thanks :heart:

addon/components/cfb-form-editor/question/default.js
```diff
@@ -0,0 +1,83 @@
...
+
+import { TYPE_MAP } from "ember-caluma/lib/field";
+
+export default class CfbFormEditorQuestionDefault extends RenderComponent {
+  @computed("model.{id,__typename}")
+  get question() {
+    const raw = {
+      ...this.model.get("data"),
+      ...this.model.get("change"),
+    };
```
anehx, line 13: Not 100% sure but I think I used `changeset.unwrap()` for this when I had the same usecase

addon/components/cfb-form-editor/question/default.js
anehx: Hm, the only idea I have is passing it down

addon/templates/components/cfb-form-editor/question/default.hbs
```diff
@@ -0,0 +1,11 @@
...
+
+{{component (get-widget this.field.question)
+  field=this.field
+  disabled=false
+  context=(hash)
+  onSave=this.onUpdate
+}}
+
+{{component hintComponent}}
+{{component errorComponent}}
```
anehx, line 11: I think you could just render the whole `cf-field` component and use the new default component just as wrapper. This way you wouldn't have to assign the errors manually

addon/components/cfb-form-editor/question/default.js
fkm-adfinis: I'm probably not completely awake yet :smile: 

I could swear that I've seen `.unwrap` somewhere lately. But I can't find it in the ember-changeset read-me :thinking: 

The idea of merging the data with the changes comes from [cfb-form-editor/question/options](https://github.com/projectcaluma/ember-caluma/blob/master/addon/components/cfb-form-editor/question/options.js#L64-L68).

addon/templates/components/cfb-form-editor/question/default.hbs
fkm-adfinis: Unfortunately, that doesn't work as expected.

Screenshot_20210105_103301
(`cf-field` is the second)

anehx: Ah, now I see the use case.. Fair enough!

addon/components/cfb-form-editor/question/default.js
anehx: Apparently it's not public API, however I think the proper way would be to use `this.model.execute()` ([docs](https://github.com/validated-changeset/validated-changeset#execute)) and then use `this.model.data`

addon/components/cfb-form-editor/question/default.js
fkm-adfinis: I found a way to access the slug. I realized that the slug is part of the URL and thus accessible via the router service.
`this.router.currentRoute.attributes.formSlug`

addon/components/cfb-form-editor/question/default.js
fkm-adfinis: Would this be save? I'm hesitant because the model is for the whole question and not just the default answer from this component. I really don't know enough about the form-builder to judge if this could have adverse effects on other parts of the form or not.

anehx: `save()` will call the underlying objects `save()` method after running `execute()`, we only want `execute()`

anehx:
@fkm-adfinis if you rebase this PR you can delete the resolutions you had to do, I resolved all those dependency lint issues in #1168

addon/components/cfb-form-editor/question/default.js
fkm-adfinis: My bad! I meant execute. As this changes the `.data` I figure this might have some consequences for `cfb-form-editor/question.js`. Or maybe the validated form that's used there. But I don't know. That's why I like the idea of not mutating `this.model`. But if you tell me that this is not a problem, I will use `raw = { ...this.model.data }` (with the execute beforehand).

fkm-adfinis:
> @fkm-adfinis if you rebase this PR you can delete the resolutions you had to do, I resolved all those dependency lint issues in #1168

Nice :sunglasses:

addon/components/cfb-form-editor/question/default.js
anehx: Hm I see your point.. I don't know what would happen, maybe try it? Otherwise keep it that way or use `unwrap()` I'll leave that up to you.

README.md
anehx, line 83: Not necessary anymore

addon/templates/components/cfb-form-editor/question.hbs
```diff
@@ -120,6 +120,17 @@
         }}
       {{/if}}
 
+      {{#if (and slug (includes f.model.__typename (array
+        "TextQuestion" "TextareaQuestion" "IntegerQuestion" "FloatQuestion"
+        "ChoiceQuestion" "MultipleChoiceQuestion" "DateQuestion" "TableQuestion"
+      )))}}
+        {{f.input
+          name="defaultAnswer"
+          label="Default value"
```
anehx, line 129: This needs to be multilingual

addon/templates/components/cfb-form-editor/question.hbs
```diff
@@ -120,6 +120,17 @@
         }}
       {{/if}}
 
+      {{#if (and slug (includes f.model.__typename (array
+        "TextQuestion" "TextareaQuestion" "IntegerQuestion" "FloatQuestion"
+        "ChoiceQuestion" "MultipleChoiceQuestion" "DateQuestion" "TableQuestion"
+      )))}}
```
anehx, line 126: I think excluding `FormQuestion` and `StaticQuestion` would be more intuitive

addon/templates/components/cfb-form-editor/question/default.hbs
```diff
@@ -0,0 +1,11 @@
+{{component labelComponent}}
+
+{{component (get-widget this.field.question)
+  field=this.field
+  disabled=false
+  context=(hash)
```
anehx, line 6: Can't this be deleted?

yarn.lock
```diff
@@ -16983,19 +16990,19 @@ uc.micro@^1.0.0, uc.micro@^1.0.1, uc.micro@^1.0.5:
...
   resolved "https://registry.yarnpkg.com/uid-number/-/uid-number-0.0.6.tgz#0ea10e8035e8eb5b8e4449f06da1c730663baa81"
   integrity sha1-DqEOgDXo61uOREnwbaHHMGY7qoE=
 
 uikit@^3.5.9:
-  version "3.6.7"
-  resolved "https://registry.yarnpkg.com/uikit/-/uikit-3.6.7.tgz#db61ef4ab282d4d2a5faa44b9226f47a97ab9124"
-  integrity sha512-cbpaqmPKvMDHdPHHsdZFoX50rI5uqMuPkoPD6wVLQ1FiJ78MMPG0RKMtqB5lGU0fcI7yzJfDX+nqaSDu/bzVsQ==
+  version "3.6.5"
+  resolved "https://registry.yarnpkg.com/uikit/-/uikit-3.6.5.tgz#384df90f96a431d12d542b227da99ceff781c71e"
+  integrity sha512-ChsoZBuCC4y0CXqJ51tiIPS4zcKgj8yvXDl8njRyMBLCjNRFlHgSjem2T6pQsEEItoN+Vh7NE3iueD70A4LW4A==
```
anehx, line 17005: Drop all changes in the lockfile, the current lockfile in master is more up to date

anehx (changes requested):

addon/templates/components/cfb-form-editor/question.hbs
fkm-adfinis: Excluding requires that you have the full list present in your mind to know the final list.

Your list is a good example as it is missing the `DynamicChoiceQuestion` and `DynamicMultipleChoiceQuestion` ;-)

addon/templates/components/cfb-form-editor/question.hbs
fkm-adfinis: Missed that. Thanks!

addon/templates/components/cfb-form-editor/question/default.hbs
fkm-adfinis: Probably, yes. I'll try and remove it if possible.

addon/templates/components/cfb-form-editor/question.hbs
anehx: Fair enough

addon/templates/components/cfb-form-editor/question/default.hbs
fkm-adfinis: Seems to work without it :sunglasses:

README.md
fkm-adfinis: Gone :)

README.md
```diff
@@ -67,6 +67,13 @@ yarn update-fragment-types
 To release a new version, head over to [Actions](https://github.com/projectcaluma/ember-caluma/actions?query=workflow%3ARelease) and trigger a workflow run.
 This will automatically check if a new version can be released, generate a changelog, draft a release and publish it on NPM.
 
+## Yarn Resolutions
+In order to solve (mostly floating) dependency issues we tend to introduce Yarn resolutions as a pragmatic and temporary solution.
+
+* **focus-trap**  
+  https://github.com/projectcaluma/ember-caluma/pull/1152  
+  Focus trap now throws an error if there is no tabbable element on initialization. This breaks the modal dialogs as they are hidden on initialization.
```
anehx, line 75: Already documented right above the resolution

addon/templates/components/cfb-form-editor/question.hbs
fkm-adfinis: Done

README.md
fkm-adfinis: I really hate those 'comments' in JSON :smile:  But I made the change.

yarn.lock
fkm-adfinis: Done (`git checkout upstream/master -- yarn.lock`)

anehx (approved):

anehx:
Tests are still failing, otherwise LGTM :100:

fkm-adfinis:
There is still a bug I found last week where removing all rows would try to save the existing rows. I'll rebase and have another look.

fkm-adfinis:
The bug is dead :sunglasses:

addon/components/cfb-form-editor/question/default.js
```diff
@@ -0,0 +1,89 @@
...
+  layout = layout;
+
+  @service router;
+
+  @computed("model.{id,__typename}")
+  get question() {
+    const raw = {
+      ...this.model.get("data"),
+      ...this.model.get("change"),
+      isRequired: "false",
```
anehx, line 19: This needs a comment

addon/templates/components/cfb-form-editor/question.hbs
anehx, line 126:
```suggestion
      {{#if (and slug (has-question-type f.model "text" "textarea" "integer" "float" "choice" "multiple-choice" "date" "table"))}}
```

anehx (changes requested):
Some minor comments, otherwise LGTM. Still needs a rebase though.

addon/components/cfb-form-editor/question/default.js
fkm-adfinis: Done. I also added a comment in `this.fields` about the structure of `this.value`.

addon/templates/components/cfb-form-editor/question.hbs
fkm-adfinis: Sexy :sunglasses: 

And done

anehx (approved):
:fire: Good work! :100:

PR #1157, opened 2020-12-15
feat(cfb): implement reordering of question options

### LINKED ISSUE #1086: Form builder: Allow users to change order of question options
Currently, question options in the form builder are appended on the bottom. This is OK for most cases, but in some places, users may want to reorder the available options (for example alphabetically, or by some other meaningful metric). The API / backend already supports this via the `sort` field, but the functionality is not exposed to the user.

Might be done via drag&drop, or some other, stupid way (like showing the "sort" field, which I wouldn't consider good UX)

Yelinz:
Allows reordering options of Choice or MultipleChoice questions, by dragging and dropping. Similar to reordering Questions.
New options aren't reorder able, as they have no slug before saving.

closes  #1086

image

### DIFF
--- a/addon/components/cfb-form-editor/question/options.js
+++ b/addon/components/cfb-form-editor/question/options.js
@@ -7,7 +7,6 @@ import lookupValidator from "ember-changeset-validations";
 import { task } from "ember-concurrency";
 import RenderComponent from "ember-validated-form/components/validated-input/-themes/uikit/render";
 import UIkit from "uikit";
-import { v4 } from "uuid";
 
 import layout from "../../../templates/components/cfb-form-editor/question/options";
 
@@ -109,6 +108,8 @@ export default RenderComponent.extend({
 
   _handleMoved({ detail: [sortable] }) {
     const options = [...sortable.$el.children];
+    // Remove last element as it is the add row button
+    options.pop();
 
     this.reorderQuestions.perform(
       options.map((option) =>
@@ -119,8 +120,6 @@ export default RenderComponent.extend({
 
   reorderQuestions: task(function* (slugs) {
     try {
-      slugs.pop();
-
       yield this.apollo.mutate({
         mutation: TYPES[this.model.__typename],
         variables: {
@@ -128,7 +127,6 @@ export default RenderComponent.extend({
             slug: this.questionSlug,
             label: this.model.label,
             options: slugs,
-            clientMutationId: v4(),
           },
         },
       });
--- a/addon/components/cfb-form-editor/question/options.js
+++ b/addon/components/cfb-form-editor/question/options.js
@@ -1,12 +1,11 @@
 import { get } from "@ember/object";
 import { reads } from "@ember/object/computed";
-import { run } from "@ember/runloop";
 import { inject as service } from "@ember/service";
+import { queryManager } from "ember-apollo-client";
 import Changeset from "ember-changeset";
 import lookupValidator from "ember-changeset-validations";
 import { task } from "ember-concurrency";
 import RenderComponent from "ember-validated-form/components/validated-input/-themes/uikit/render";
-import UIkit from "uikit";
 
 import layout from "../../../templates/components/cfb-form-editor/question/options";
 
@@ -33,7 +32,7 @@ export default RenderComponent.extend({
 
   intl: service(),
   notification: service(),
-  apollo: service(),
+  apollo: queryManager(),
 
   questionSlug: reads("model.slug"),
 
@@ -68,11 +67,6 @@ export default RenderComponent.extend({
     );
   },
 
-  didInsertElement(...args) {
-    this._super(...args);
-    this.setupUIkit();
-  },
-
   _update() {
     this.update({
       edges: this.optionRows
@@ -98,23 +92,6 @@ export default RenderComponent.extend({
     });
 
     this.setDirty();
-  },
-
-  setupUIkit() {
-    UIkit.util.on("#option-list", "moved", (...args) =>
-      run(this, this._handleMoved, ...args)
-    );
-  },
-
-  _handleMoved({ detail: [sortable] }) {
-    // Remove last element as it is the add row button
-    const options = [...sortable.$el.children].slice(0, -1);
-
-    this.reorderOptions.perform(
-      options.map((option) =>
-        addQuestionPrefix(option.firstElementChild.id, this.questionSlug)
-      )
-    );
   },
 
   reorderOptions: task(function* (slugs) {
@@ -191,6 +168,17 @@ export default RenderComponent.extend({
 
     update() {
       this._update();
+    },
+
+    _handleMoved({ detail: [sortable] }) {
+      // Remove last element as it is the add row button
+      const options = [...sortable.$el.children].slice(0, -1);
+
+      this.reorderOptions.perform(
+        options.map((option) =>
+          addQuestionPrefix(option.firstElementChild.id, this.questionSlug)
+        )
+      );
     },
   },
 });
--- a/addon/components/cfb-form-editor/question/options.js
+++ b/addon/components/cfb-form-editor/question/options.js
@@ -1,14 +1,25 @@
 import { get } from "@ember/object";
 import { reads } from "@ember/object/computed";
+import { run } from "@ember/runloop";
 import { inject as service } from "@ember/service";
 import Changeset from "ember-changeset";
 import lookupValidator from "ember-changeset-validations";
+import { task } from "ember-concurrency";
 import RenderComponent from "ember-validated-form/components/validated-input/-themes/uikit/render";
+import UIkit from "uikit";
+import { v4 } from "uuid";
 
 import layout from "../../../templates/components/cfb-form-editor/question/options";
 
+import saveChoiceQuestionMutation from "ember-caluma/gql/mutations/save-choice-question";
+import saveMultipleChoiceQuestionMutation from "ember-caluma/gql/mutations/save-multiple-choice-question";
 import slugify from "ember-caluma/utils/slugify";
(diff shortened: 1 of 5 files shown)


addon/components/cfb-form-editor/question/options.js
```diff
@@ -83,6 +101,52 @@ export default RenderComponent.extend({
...
+      slugs.pop();
+
+      yield this.apollo.mutate({
+        mutation: TYPES[this.model.__typename],
+        variables: {
+          input: {
+            slug: this.questionSlug,
+            label: this.model.label,
+            options: slugs,
+            clientMutationId: v4(),
```
czosel, line 131: Out of curiosity: I see that we're setting `clientMutationId` also in other mutations - why? I thought this was optional.

czosel (approved):
Code LGTM

addon/components/cfb-form-editor/question/options.js
anehx: Isn't necessary.. Drop it.

addon/components/cfb-form-editor/question/options.js
```diff
@@ -83,6 +101,52 @@ export default RenderComponent.extend({
...
+    this.reorderQuestions.perform(
+      options.map((option) =>
+        addQuestionPrefix(option.firstElementChild.id, this.questionSlug)
+      )
+    );
+  },
+
+  reorderQuestions: task(function* (slugs) {
+    try {
+      slugs.pop();
```
anehx, line 122: Why this? Also, I'd prefer a functional approach like `const [ slugsWithoutLast, _ ] = slugs;`. A comment why this is done would be nice as well

addon/templates/components/cfb-form-editor/question/options.hbs
```diff
@@ -1,13 +1,21 @@
...
+<ul class="uk-list uk-list-divider uk-margin-remove-bottom uk-margin-small-top" id="option-list" uk-sortable="handle: .uk-sortable-handle;">
   {{#each optionRows as |row i|}}
     {{#validated-form
       data-test-row=(concat "option-" (add i 1))
       tagName="li"
       model=row
     as |f|}}
-      <div uk-grid class="uk-grid-small uk-flex">
+      <div uk-grid class="uk-grid-small uk-flex uk-flex-middle" id={{row.slug}}>
+        {{#if (not (and row.isNew (gt optionRows.length 1)))}}
```
anehx, line 11:
```suggestion
        {{#if (not (and row.isNew (gt this.optionRows.length 1)))}}
```

anehx (changes requested):

addon/components/cfb-form-editor/question/options.js
Yelinz: Removed it

addon/components/cfb-form-editor/question/options.js
Yelinz: Moved it and added a comment.
The suggestion you wrote doesnt work as it would only take the first 2 elements from the array and you cant reverse the rest operator.

addon/templates/components/cfb-form-editor/question/options.hbs
Yelinz: Changed

addon/components/cfb-form-editor/question/options.js
```diff
@@ -83,6 +100,51 @@ export default RenderComponent.extend({
...
+  setupUIkit() {
+    UIkit.util.on("#option-list", "moved", (...args) =>
+      run(this, this._handleMoved, ...args)
+    );
+  },
+
+  _handleMoved({ detail: [sortable] }) {
+    const options = [...sortable.$el.children];
+    // Remove last element as it is the add row button
+    options.pop();
```
anehx, line 112:
```suggestion
    // Remove last element as it is the add row button
    const options = [...sortable.$el.children].slice(0, -1);
```
Is easier to understand and doesn't mutate the `options`

addon/components/cfb-form-editor/question/options.js
```diff
@@ -83,6 +100,51 @@ export default RenderComponent.extend({
...
+
+    this.reorderQuestions.perform(
+      options.map((option) =>
+        addQuestionPrefix(option.firstElementChild.id, this.questionSlug)
+      )
+    );
+  },
+
+  reorderQuestions: task(function* (slugs) {
```
anehx, line 121: Shouldn't that be `reorderOptions`?

addon/components/cfb-form-editor/question/options.js
```diff
@@ -83,6 +100,51 @@ export default RenderComponent.extend({
     this.setDirty();
   },
 
+  setupUIkit() {
+    UIkit.util.on("#option-list", "moved", (...args) =>
+      run(this, this._handleMoved, ...args)
+    );
+  },
```
anehx, line 107: I know it's very convenient to solve this that way since you can just copy paste from the reorder question functionality. However the modern / new way to solve this would be a custom modifier. You should ask the person that gave you the assignment which way they'd prefer.

anehx (changes requested):

addon/components/cfb-form-editor/question/options.js
Yelinz: Renamed

addon/components/cfb-form-editor/question/options.js
Yelinz: Changed

addon/components/cfb-form-editor/question/options.js
Yelinz: To rewrite this using modifieres and still have it looking similar or as good as uikit would take a good amount of time, so this could just be an future issue of migrating to modifiers

anehx: Huh? I meant a modifier that uses UIkit.

anehx: E.g:

```hbs
<ul
  class="uk-list uk-list-divider uk-margin-remove-bottom uk-margin-small-top"
  id="option-list"
  {{uk-sortable @handle=".uk-sortable-handle" @onMoved=this._handleMoved}}
>
```

anehx: It's just that we write code that has to be refactored to 100% since this thing here is not possible with glimmer components. I don't like throwaway code.

Yelinz: I thought you meant to use https://github.com/ember-modifier/ember-modifier

Changed to use the uk-sortable component from ember-uikit

anehx: Well I did mean that. My example is exactly that.

anehx: However the sortable component is ok as well..

addon/templates/components/cfb-form-editor/question/options.hbs
```diff
@@ -1,15 +1,29 @@
 {{component labelComponent}}
 
-<ul class="uk-list uk-list-divider uk-margin-remove-bottom uk-margin-small-top">
-  {{#each optionRows as |row i|}}
+{{#uk-sortable
```
anehx, line 3: Angle bracket notation please

addon/templates/components/cfb-form-editor/question/options.hbs
```diff
@@ -1,15 +1,29 @@
...
+  threshold=".uk-sortable-handle"
```
anehx, line 4: I'm 100% sure this won't work - did you test it??

addon/templates/components/cfb-form-editor/question/options.hbs
```diff
@@ -1,15 +1,29 @@
...
+  on-moved=(action "_handleMoved")
+  class="uk-list uk-list-divider uk-margin-remove-bottom uk-margin-small-top"
+  id="option-list"
```
anehx, line 7: Why?

Yelinz: Just copy pasted, didnt seem necessary so removedit

addon/templates/components/cfb-form-editor/question/options.hbs
Yelinz: Yes tested it and it works

addon/templates/components/cfb-form-editor/question/options.hbs
Yelinz: Changed

addon/templates/components/cfb-form-editor/question/options.hbs
anehx: Pretty sure what you're passing there is supposed to be the `handle`. `threshold` is a number - there's a typo in the documentation that might've caused this error here but this certainly doesn't work.

Yelinz: Changed it to handle

addon/components/cfb-form-editor/question/options.js
```diff
@@ -22,6 +30,8 @@ export default RenderComponent.extend({
   layout,
 
   intl: service(),
+  notification: service(),
+  apollo: service(),
```
anehx, line 34: This needs to be the `queryManager` by `ember-apollo-client`: https://github.com/ember-graphql/ember-apollo-client/tree/v2.2.1#usage

anehx:
After [this](https://github.com/projectcaluma/ember-caluma/pull/1157#discussion_r561929687) we're good to go!

addon/components/cfb-form-editor/question/options.js
Yelinz: Changed

anehx (approved):

PR #1168, opened 2021-01-05
chore(lint): fix ember-cli-dependency-lint config

anehx:
The configuration needs to be in the dummy app config folder, see salsify/ember-cli-dependency-lint#16. Also, the configuration of ember-changeset and ember-changeset-validations was obsolete since we updated this a while ago.


anehx:
This PR also removes all resolutions and updates the packages needed for dependency lint to pass

anehx:
Oh, almost forgot why this is relevant: the configuration should configure that the dependency linter is **not** run as a test in the test suite but a separate command. When I configured this initially I was assuming that it works - however it didn't. This caused the linter to be run with floating dependencies which makes absolutely zero sense and broke our CI all the time.

velrest (approved):

package.json
```diff
@@ -125,6 +125,12 @@
     "typeface-oxygen": "1.1.13",
     "typeface-oxygen-mono": "1.1.13"
   },
+  "//":{
+    "focus-trap": "Focus trap now an error if there is no tabbable element on initialization. This breaks the modal dialogs as they are hidden on initialization."
```
velrest, line 129:
```suggestion
    "focus-trap": "Focus trap now throws an error if there is no tabbable element on initialization. This breaks the modal dialogs as they are hidden on initialization."
```

PR #1173, opened 2021-01-06
feat(form-builder): add calculated fields for the form builder

anehx:


addon/components/cfb-form-editor/question.js
```diff
@@ -292,26 +299,18 @@ export default Component.extend({
...
-                meta: JSON.stringify(changeset.get("meta").unwrap()),
-                isArchived: changeset.get("isArchived"),
-                clientMutationId: v4(),
-              },
-              this[`_get${changeset.get("__typename")}Input`](changeset)
-            ),
+            input: {
+              ...this._getBaseInput(changeset),
+              ...(
+                this[`_get${changeset.get("__typename")}Input`] || (() => ({}))
```
czosel, line 311: :laughing: this line was  already funny before, and this change makes it even better!

For the more casual reader it might be more friendly to use a variable assignment though.

addon/helpers/has-question-type.js
```diff
@@ -0,0 +1,14 @@
...
+import { dasherize } from "@ember/string";
+
+export default helper(function hasQuestionType(
+  [obj, ...expected],
+  { replace = "Question" }
+) {
+  const typename = dasherize((obj.__typename || "").replace(replace, ""));
+  const typenames = expected.map((typename) =>
+    dasherize(typename.replace(replace, ""))
+  );
```
czosel, line 11: Nitpick: This might become a little more readable if you extract the `dasherize(... replace())` into a little helper function.

czosel:
Apart from the nitpicks, this looks great :+1:

addon/helpers/has-question-type.js
anehx: `dasherizeAndReplace` or what? I really don't think that would help at all... This is very basic string alternation, shouldn't be too complicated for anyone to understand.

addon/components/cfb-form-editor/question.js
anehx: Fair

addon/helpers/has-question-type.js
anehx: As discussed, done

addon/components/cfb-form-editor/question.js
anehx: Done

czosel (approved):

PR #1174, opened 2021-01-06
feat(form): show calculated fields in form and compute their value live

anehx:
This commit also refactors the float input and label components to glimmer components since I had to touch them.

Depends on #1173

### DIFF
--- a/addon/lib/field.js
+++ b/addon/lib/field.js
@@ -226,7 +226,11 @@ export default Base.extend({
    * @property {Boolean} isDirty
    * @accessor
    */
-  isDirty: computed("isNew", "isDefault", function () {
+  isDirty: computed("question.isCalculated", "isNew", "isDefault", function () {
+    if (this.question.isCalculated) {
+      return false;
+    }
+
     return !this.isNew && !this.isDefault;
   }),
 
@@ -252,7 +256,48 @@ export default Base.extend({
    * @property {*} value
    * @accessor
    */
-  value: reads("answer.value"),
+  value: computed(
+    "question.isCalculated",
+    "calculatedValue",
+    "answer.value",
+    function () {
+      if (this.question.isCalculated) {
+        return this.calculatedValue;
+      }
+
+      return this.answer.value;
+    }
+  ),
+
+  /**
+   * The computed value of a caluclated question
+   *
+   * @property {*} calculatedValue
+   * @accessor
+   */
+  calculatedValue: computed(
+    "document.jexl",
+    "calculatedDependencies.@each.{hidden,value}",
+    "jexlContext",
+    "question.{isCalculated,calcExpression}",
+    function () {
+      if (
+        !this.question.isCalculated ||
+        !this.calculatedDependencies.every((field) => !field.hidden)
+      ) {
+        return null;
+      }
+
+      try {
+        return this.document.jexl.evalSync(
+          this.question.calcExpression,
+          this.jexlContext
+        );
+      } catch (error) {
+        return null;
+      }
+    }
+  ),
 
   /**
    * Fetch all formerly used dynamic options for this question. This will be
@@ -408,6 +453,39 @@ export default Base.extend({
   ),
 
   /**
+   * Fields that are referenced in the `calcExpression` JEXL expression
+   *
+   * If the value or hidden state of any of these fields change, the JEXL
+   * expression needs to be re-evaluated.
+   *
+   * @property {Field[]} calculatedDependencies
+   * @accessor
+   */
+  calculatedDependencies: computed(
+    "document.{jexl,fields.[]}",
+    "question.{isCalculated,calcExpression}",
+    function () {
+      if (!this.question.isCalculated) {
+        return [];
+      }
+
+      return getDependenciesFromJexl(
+        this.document.jexl,
+        this.question.calcExpression
+      ).map((slug) => {
+        const f = this.document.findField(slug);
+
+        assert(
+          `Field for question \`${slug}\` was not found in this document. Please check the \`calcExpression\` jexl expression: \`${this.question.calcExpression}\`.`,
+          f
+        );
+
+        return f;
+      });
+    }
+  ),
+
+  /**
    * Fields that are referenced in the `isHidden` JEXL expression
    *
    * If the value or hidden state of any of these fields change, the JEXL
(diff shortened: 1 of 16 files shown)


addon/lib/field.js
```diff
@@ -407,6 +444,39 @@ export default Base.extend({
...
+
+        assert(
+          `Field for question \`${slug}\` was not found in this document. Please check the \`calcExpression\` jexl expression: \`${this.question.calcExpression}\`.`,
+          f
+        );
+
+        return f;
+      });
+    }
+  ),
```
czosel, line 478: Did you already check if this has a noticable performance impact in larger forms?

addon/lib/field.js
```diff
@@ -252,7 +256,40 @@ export default Base.extend({
...
+  calculatedValue: computed(
+    "document.jexl",
+    "calculatedDependencies.@each.{hidden,value}",
+    "jexlContext",
+    "question.{isCalculated,calcExpression}",
+    function () {
+      if (
+        !this.question.isCalculated ||
+        !this.calculatedDependencies.every(
+          (field) => !field.hidden && typeof field.value === "number"
```
czosel, line 281: Is it really a requirement that calculated fields are only working with numbers? I don't have a concrete use-case, but maybe we'd like to work with strings as well? Would that cause trouble anywhere?

addon/lib/field.js
anehx: Not sure what you'd expect TBH. If you don't add like 20 fields with formulas using 20 different fields I'm not sure how this would impact the performance

addon/lib/field.js
anehx: Not sure what the backend does here. However, I don't think that string concatenation is a valid use case for a caluma field. We should keep it as simple as possible without allowing stuff that could cause problems if we don't have a use case for it

czosel: cc @sbor23: do we have this limitation in the backend as well? If we decide to enforce this limitation, I'd prefer to do it alreading (additionally) in `SaveCalculatedFloatQuestion`. I currently don't really see how handling strings as well could cause problems.

addon/lib/field.js
czosel: I didn't think too deeply about this, I just remember that we (or rather you :wink) optimized this code a bit in the past and computed properties can sometimes cause an overhead. I just wanted to bring this up, we can also check the performance side of things after the PR landed.

anehx: It won't be a problem. There will be very few fields with content in this. Also, I optimized that dependency logic to exactly this (computed properties) because it the most performant. This will be ok.

addon/lib/field.js
anehx: I still don't think we should support strings since there is no use case for this.. @kaldras what do you think?

addon/lib/field.js
czosel: Alright :+1:

addon/lib/field.js
czosel: I think I found the relevant backend code snippet (validators.py):
```py
        question_jexl = jexl.QuestionJexl()
        deps = set(question_jexl.extract_referenced_questions(expr))
        illegal_deps = ", ".join(
            models.Question.objects.filter(
                pk__in=deps, type=models.Question.TYPE_CALCULATED_FLOAT
            ).values_list("slug", flat=True)
        )

        if illegal_deps:
            raise exceptions.ValidationError(
                f"Calc expression references other calculated question: {illegal_deps}"
            )
```

sbor23: The backend doesn't do anything with the input types nor the output types. However we went for a "typed" GraphQL schema for calculated fields. It returns a `FloatAnswer` so we could enforce this more strictly in the future. But theoretically a String could be returned even right now, but I don't know how graphql (frontend or backend) would handle such a case.

kaldras: If I understand correctly the backend would try to evaluate the jexl even for string answers right now but the frontend does not?

I don't see a use case for other types than numbers/integers right now, so I think only accessing number answers would be ok. However it is done, is should be done the same way in the front and backend right?

sbor23: @czosel yes the backend doesn't validate anything except that there are no referenced calculated questions. 

> If I understand correctly the backend would try to evaluate the jexl even for string answers right now but the frontend does not?

For the backend: yes

> I don't see a use case for other types than numbers/integers right now, so I think only accessing number answers would be ok. However it is done, is should be done the same way in the front and backend right?

Probably yeah.

anehx: We should only allow integer and float questions as valid dependencies since we are talking about a Calculated**Float**Question which has to return a float. Strings don't make any sense.

czosel: Oh right - I totally missed the fact that it would be quite weird to build a float from strings. @sbor23, can you make the backend validation more strict as @anehx suggested?

czosel: I'll go ahead and resolve this, because the frontend implementation is good :upside_down_face:

czosel (approved):

czosel:
Awesome work @anehx btw! :100:

tests/unit/lib/field-test.js
```diff
@@ -437,4 +437,32 @@ module("Unit | Library | field", function (hooks) {
...
+  module("calculated", function (hooks) {
+    hooks.beforeEach(function () {
+      this.calculated = this.document.findField("calculated");
+      this.float = this.document.findField("float");
+    });
+
+    test("calculates the correct value", async function (assert) {
+      assert.expect(1);
+
+      // 1.1 + 5 + 100 = 501.1
```
sbor23, line 450: `1.1 + 5 * 100 = 501.1`

addon/lib/field.js
sbor23: > We should only allow integer and float questions as valid dependencies since we are talking about a Calculated**Float**Question which has to return a float. Strings don't make any sense.

This is a wrong assumption IMO.
A valid calcExpression could be something like `'my-text-question'|answer in ['foo', 'bar'] ? 1 : 0`. So `my-text-question` could obviously be a `TextQuestion` (or `StringQuestion`...). So there is no way you can validate the input types without doing actual static analysis on it.

czosel: Good point, didn't think of the ternary idea.

czosel: LGTM now

PR #1193, opened 2021-01-21
fix(cfb): fix recalculation issue in default values

fkm-adfinis:
This adds a decorator to the field getter to stop Ember from recalculating the property too often.

`@computed("question", "value", "value.content", "model.slug")`
would technically be enough. The other properties are there to satisfy a linting rule.


kaldras (approved):

anehx:
[Caching](https://guides.emberjs.com/release/in-depth-topics/autotracking-in-depth/#toc_caching-of-tracked-properties) (there's a polyfill) might also be an option for such things. Getters will be entered multiple times even when no dependent property changed - it's their nature.

PR #1197, opened 2021-01-22
fix(form): add fields used in mapby transforms to dependencies

anehx:


fkm-adfinis (approved):

addon/lib/dependencies.js
```diff
@@ -0,0 +1,134 @@
...
+  const allTransforms = getTransforms(getAST(jexl, expression));
+  const answerTransforms = allTransforms.filter(
+    (transform) => transform.name === "answer"
+  );
+  const mapbyTransforms = allTransforms.filter(
+    (transform) =>
+      transform.name === "mapby" &&
+      transform.args[0].type === "FunctionCall" &&
+      transform.args[0].name === "answer"
+  );
```
czosel, line 26: I tried following this using the JEXL playground: [see here](https://czosel.github.io/jexl-playground/#/?input='bar'%7Canswer%7Cmapby('column')&transforms=%7B%0A%20%20mapby%3A%20(name%2C%20col)%20%3D%3E%20name%20%2B%20%22col%3A%22%20%2Bcol%2C%0A%20%20answer%3A%20name%20%3D%3E%20%22Answer%20%22%20%2B%20name%2C%0A%7D).

However it seems that the AST structure looks a little different: The "mapby" only has the literal string as argument, and the part in front needs to be accessed via `subject`. Could it be that this changed recently in JEXL, or am I missing something else?

Input
```
'bar'|answer|mapby('column')
```

AST
```
{
  "type": "Transform",
  "name": "mapby",
  "args": [
    {
      "type": "Literal",
      "value": "column"
    }
  ],
  "subject": {
    "type": "Transform",
    "name": "answer",
    "args": [],
    "subject": {
      "type": "Literal",
      "value": "bar"
    }
  }
}
```

czosel: Indeed, the playground needs an update. The AST in 2.3.0 is

```
{
  "type": "FunctionCall",
  "name": "mapby",
  "args": [
    {
      "type": "FunctionCall",
      "name": "answer",
      "args": [
        {
          "type": "Literal",
          "value": "bar"
        }
      ],
      "pool": "transforms"
    },
    {
      "type": "Literal",
      "value": "column"
    }
  ],
  "pool": "transforms"
}
```

czosel: JEXL Playground updated :sparkles:

addon/lib/dependencies.js
```diff
@@ -0,0 +1,134 @@
...
+export function getDependenciesFromJexl(jexl, expression) {
+  const allTransforms = getTransforms(getAST(jexl, expression));
+  const answerTransforms = allTransforms.filter(
+    (transform) => transform.name === "answer"
+  );
+  const mapbyTransforms = allTransforms.filter(
+    (transform) =>
+      transform.name === "mapby" &&
+      transform.args[0].type === "FunctionCall" &&
+      transform.args[0].name === "answer"
```
czosel, line 25: Should we also make sure that `transform.args[0].args[0].type == "Literal"`? Otherwise e.g. `bar|answer|mapby('column')` would cause a dependency to `bar` to be listed. Since we don't support this we might want to throw an error or a warning in that case?

czosel: If we decide to do that, we should do the same for the `answer` transform.

addon/lib/dependencies.js
```diff
@@ -0,0 +1,134 @@
...
+ */
+export function findField(document, slug, expression) {
+  const field = document.findField(slug);
+
+  assert(
+    `Field for question \`${slug}\` could not be found in the document \`${document.uuid}\`. Please verify that the jexl expression is correct: \`${expression}\`.`,
+    field
+  );
+
+  return field;
```
czosel, line 55: Are there cases where we're calling `document.findField` with non-existing slugs but don't want the error? If not, I'd move the assertion into `document.findField` and drop this method.

czosel:
Looks great apart from the nitpicks, awesome work! :100:

addon/lib/dependencies.js
anehx: That can't happen since the backend validates this..

addon/lib/dependencies.js
anehx: That would be a breaking change since `findField` is public API that is used in various projects. They depend on it returning undefined if the field is not found. Asserting something there is not a good idea IMO

czosel: :+1: Fair enough!

addon/lib/dependencies.js
czosel: :+1: Alright

czosel (approved):

PR #1202, opened 2021-01-26
fix(form-builder): fix fieldset on generated field for the default value

anehx:


aziiee (approved):
Thanks for fix! Please fix the failing tests so we can merge this

PR #1227, opened 2021-02-10
fix(lib/navigation): don't set state to empty if fields are required

fkm-adfinis:
Otherwise, there's no other way than do go though all (sub-)forms
to check where the missing question is located.

### DIFF
--- a/addon/lib/field.js
+++ b/addon/lib/field.js
@@ -219,13 +219,19 @@ export default Base.extend({
    * @property {Boolean} isDirty
    * @accessor
    */
-  isDirty: computed("question.isCalculated", "isNew", "isDefault", function () {
-    if (this.question.isCalculated) {
-      return false;
-    }
+  isDirty: computed(
+    "isDefault",
+    "isNew",
+    "question.isCalculated",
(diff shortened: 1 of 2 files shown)


fkm-adfinis:
Screenshot_20210210_121942

The question marks show where there are missing questions. The orange color shows that they are dirty.

anehx (changes requested):
That doesn't make any sense to me.. Why are we excluding empty optional field? This would mean that when the user only fills an optional field the state in the navigation would still show empty - which is wrong.

fkm-adfinis:
We are excluding non-optional fields to help the user find forms with missing fields that are required.

The `empty`/`in-progress` might be used incorrectly here. But I'm open to suggestions. What I need is for the navigation to show which form is missing input. Even if the form has never been touched.

anehx:
LGTM, if you think you tested it and it does what it should you can go ahead and merge.

anehx (approved):

anehx:
Ah, commit message now doesn't fit the changes anymore - please fix that before merging..

PR #1230, opened 2021-02-10
feat(jexl): add transforms for mathematical functions

aziiee:


anehx (changes requested):
Tests are missing

czosel:
While we're at it, I'd consider the following functions useful as well: `pow`, `abs`. For `round`, it might be convenient to expose a second argument to specify the number of decimal places that should be returned (i.e. `(num, places) => Math.round(num) * Math.pow(10, places) / Math.pow(10, places)`).

A last remark which is probably out of scope for now but might become interesting in the future: Should we consider function syntax as well? (i.e. `pow(2, 3)` instead of `2|pow(3)`)

anehx:
> While we're at it, I'd consider the following functions useful as well: `pow`, `abs`. For `round`, it might be convenient to expose a second argument to specify the number of decimal places that should be returned (i.e. `(num, places) => Math.round(num) * Math.pow(10, places) / Math.pow(10, places)`).
> 
> A last remark which is probably out of scope for now but might become interesting in the future: Should we consider function syntax as well? (i.e. `pow(2, 3)` instead of `2|pow(3)`)

I agree that `round` should accept a parameter for digits - this needs to be done. However, I don't think `pow` and `abs` are really useful at the moment. `rounddown` (`Math.floor`) and `roundup` (`Math.ceil`) would be more helpful at the moment.

addon/lib/document.js
```diff
@@ -127,6 +127,9 @@ export default Base.extend({
       console.debug(`${label}:`, any);
       return any;
     });
+    documentJexl.addTransform("min", (arr) => Math.min(...arr));
+    documentJexl.addTransform("max", (arr) => Math.max(...arr));
+    documentJexl.addTransform("round", (num) => Math.round(num));
```
anehx, line 132: You should handle what happens if there is a non number as input. I don't think you want `NaN` in your JEXL. Those should return `null` when failing as do the other transforms.

aziiee:
> > While we're at it, I'd consider the following functions useful as well: `pow`, `abs`. For `round`, it might be convenient to expose a second argument to specify the number of decimal places that should be returned (i.e. `(num, places) => Math.round(num) * Math.pow(10, places) / Math.pow(10, places)`).
> > A last remark which is probably out of scope for now but might become interesting in the future: Should we consider function syntax as well? (i.e. `pow(2, 3)` instead of `2|pow(3)`)
> 
> I agree that `round` should accept a parameter for digits - this needs to be done. However, I don't think `pow` and `abs` are really useful at the moment. `rounddown` (`Math.floor`) and `roundup` (`Math.ceil`) would be more helpful at the moment.

I'll make the following changes in that case:
- Parameter for `Math.round()`
- Add `Math.ceil()`, `Math.floor()`

sbor23:
Just to ensure the same behavior in both js and python.
In pyjexl the args are optional by default, meaning that `1.5|round" == 2`, `1.5|round() == 2.0`, `"1.55"|round(1) == 1.6`.
Can we expect the same from js jexl?

anehx:
> Just to ensure the same behavior in both js and python.
> In pyjexl the args are optional by default, meaning that `1.5|round == 2`, `1.5|round() == 2.0`, `1.55|round(1) == 1.6`.
> Can we expect the same from js jexl?

It has to be implemented that way, yes @aziiee

czosel:
@sbor23 Python compat is a good point:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round#description
> If the fractional portion is exactly 0.5, the argument is rounded to the next integer in the direction of +∞. Note that this differs from many languages' round() functions, which often round this case to the next integer away from zero, instead giving a different result in the case of negative numbers with a fractional part of exactly 0.5.

JS: `Math.round(0.5) == 1`
Py: `round(0.5) == 0`

czosel:
Follow-up: For the python implementation, we should use the `decimal` module with the ROUND_CEILING strategy.
https://docs.python.org/3/library/decimal.html#rounding-modes

addon/lib/document.js
aziiee: Updated as requested

addon/lib/document.js
```diff
@@ -127,6 +127,21 @@ export default Base.extend({
       console.debug(`${label}:`, any);
       return any;
     });
+    documentJexl.addTransform("min", (arr) => Math.min(...arr));
+    documentJexl.addTransform("max", (arr) => Math.max(...arr));
+    documentJexl.addTransform("round", (num, places) =>
+      isNaN(num)
+        ? null
+        : !isNaN(places)
```
anehx, line 135: Why? This isn't user input. We can assume that this is a number

anehx (changes requested):

addon/lib/document.js
aziiee: That is true, fixed

addon/lib/document.js
```diff
@@ -127,6 +127,21 @@ export default Base.extend({
       console.debug(`${label}:`, any);
       return any;
     });
+    documentJexl.addTransform("min", (arr) => Math.min(...arr));
+    documentJexl.addTransform("max", (arr) => Math.max(...arr));
```
anehx, line 131: Those `arr`s need to be filtered, including a test for it

sbor23:
> @sbor23 Python compat is a good point:
> 
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round#description
> 
> > If the fractional portion is exactly 0.5, the argument is rounded to the next integer in the direction of +∞. Note that this differs from many languages' round() functions, which often round this case to the next integer away from zero, instead giving a different result in the case of negative numbers with a fractional part of exactly 0.5.
> 
> JS: `Math.round(0.5) == 1`
> Py: `round(0.5) == 0`

Damn I wasn't even aware of the python3 default implementation! :see_no_evil:

> Follow-up: For the python implementation, we should use the `decimal` module with the ROUND_CEILING strategy.
> https://docs.python.org/3/library/decimal.html#rounding-modes

Well let's not turn this into an argument which is the "right" rounding strategy, as the js-way is definitely the most intuitive and "taught in school". So this is the way as you proposed.

anehx:
> Well let's not turn this into an argument which is the "right" rounding strategy, as the js-way is definitely the most intuitive and "taught in school". So this is the way as you proposed.

Haha, exactly what I said as well :wink: Totally agree!

addon/lib/document.js
anehx: Also, the backend implemented a `sum` function (totally useful transform!) so the JS should implement this as well.

```js
documentJexl.addTransform("sub", (arr) => {
  const nums = arr.filter((num) => typeof num === "number");
  return nums.reduce((num, base) => base + num, 0);
});

Also maybe we should add `avg` for the average?
```

addon/lib/document.js
```diff
@@ -127,6 +127,30 @@ export default Base.extend({
       console.debug(`${label}:`, any);
       return any;
     });
+    documentJexl.addTransform("min", (arr) => {
+      const nums = arr.filter((num) => !isNaN(num));
+      return nums.length ? Math.min(...nums) : null;
+    });
+    documentJexl.addTransform("max", (arr) => {
+      const nums = arr.filter((num) => !isNaN(num));
```
anehx, line 135: Maybe extract this so it's not duplicated? Something like `const onlyNumbers = (nums) => nums.filter((num) => !isNaN(num));`

addon/lib/document.js
aziiee: Agree, I've added both since we can benefit from both of them

anehx (changes requested):

addon/lib/document.js
aziiee: Done!

addon/lib/document.js
```diff
@@ -116,6 +116,7 @@ export default Base.extend({
    */
   jexl: computed(function () {
     const documentJexl = new jexl.Jexl();
+    const onlyNumbers = (nums) => nums.filter((num) => !isNaN(num));
```
anehx, line 119: Take this out of this property to the top of the file. This way the function is redefined in each computation of this property..

addon/lib/document.js
```diff
@@ -127,6 +128,34 @@ export default Base.extend({
...
+        : Math.round(num)
+    );
+    documentJexl.addTransform("ceil", (num) =>
+      isNaN(num) ? null : Math.ceil(num)
+    );
+    documentJexl.addTransform("floor", (num) =>
+      isNaN(num) ? null : Math.floor(num)
+    );
+    documentJexl.addTransform("sum", (arr) =>
+      onlyNumbers(arr).reduce((num, base) => base + num, 0)
```
anehx, line 153: Extract `.reduce((num, base) => base + num, 0)` into a `sum` function since you use that in the `avg` function as well

addon/lib/document.js
```diff
@@ -127,6 +128,34 @@ export default Base.extend({
...
+    });
+    documentJexl.addTransform("max", (arr) => {
+      const nums = onlyNumbers(arr);
+      return nums.length ? Math.max(...nums) : null;
+    });
+    documentJexl.addTransform("round", (num, places) =>
+      isNaN(num)
+        ? null
+        : places
+        ? Math.round(num * Math.pow(10, places)) / Math.pow(10, places)
```
anehx, line 143: Drop the line below and provide a default value of `0` for the argument `places`. `Math.pow(10, 0)` will result in `1` so it does the right thing. This way we can get rid of the nested ternary which is really hard to read

anehx (changes requested):

anehx (approved):
Looking good! Sorry for the many nitpicks.. Good work! :fire:

PR #1233, opened 2021-02-11
fix(jexl): fix division by 0 in avg transform

aziiee:
Calculating the average of an empty array would result in a `0/0`
division. `null` has to be returned in that case instead of `NaN`.


aziiee:
cc @sbor23

anehx (approved):

PR #1234, opened 2021-02-12
fix(jexl): fix validation of numbers in transforms

aziiee:
We need to validate numbers differently since `null` `""` `true`
and `false` are not `NaN` and therefore a validation with `isNaN()`
is pretty much useless.

### DIFF
--- a/addon/lib/document.js
+++ b/addon/lib/document.js
@@ -8,7 +8,8 @@ import { decodeId } from "ember-caluma/helpers/decode-id";
 import Base from "ember-caluma/lib/base";
 import { intersects } from "ember-caluma/utils/jexl";
 
-const onlyNumbers = (nums) => nums.filter((num) => !isNaN(num));
+const onlyNumbers = (nums) =>
+  nums.filter((num) => !isNaN(num) && typeof num === "number");
 const sum = (nums) => nums.reduce((num, base) => base + num, 0);
 
 /**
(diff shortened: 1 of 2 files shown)


aziiee:
cc @sbor23

sbor23 (approved):
Seems to be the better solution, a quick test with floats works too. But I'm not a JS expert so I'd prefer some frontend people to look at it too.

Edit: Testing could still be a bit more extensive, e.g. I tested `onlyNumber([1, 2, "3", "1", "nan", NaN, null, 1.2, 1.0])` because not some was missing in tests...

addon/lib/document.js
```diff
@@ -8,7 +8,10 @@ import { decodeId } from "ember-caluma/helpers/decode-id";
 import Base from "ember-caluma/lib/base";
 import { intersects } from "ember-caluma/utils/jexl";
 
-const onlyNumbers = (nums) => nums.filter((num) => !isNaN(num));
+const onlyNumbers = (nums) =>
+  nums.filter(
+    (num) => Number.isInteger(num) || (Number(num) === num && num % 1 !== 0)
```
anehx, line 13: Uhm, how about `(num) => typeof num === "number"`? I thought we agreed that we don't want to cast the input.

anehx (changes requested):

addon/lib/document.js
sbor23: It don't know exactly what the right part of the OR does, but it doesn't allow strings. My test above yielded this result: `[ 1, 2, 1.2, 1 ]`

aziiee: As discussed with @kaldras, I've updated it to `(num) => !isNaN(num) && typeof num === "number"` since `typeof NaN === "number" results in `true``

kaldras (approved):

PR #1258, opened 2021-02-24
feat(cfb): implement search and pagination of form list

luytena:
Add the capability of searching for specific form names among the
listed forms. Paginate the retrieved forms and allow to load more
forms on demand.

### DIFF
--- a/addon/caluma-query/models/form.js
+++ b/addon/caluma-query/models/form.js
@@ -1,0 +1,11 @@
+import CalumaQueryModel from "./index";
+
+export default class FormModel extends CalumaQueryModel {
+  static fragment = `{
+    name
+    slug
+    description
+    isArchived
+    isPublished
+  }`;
+}
--- a/addon/caluma-query/queries/form.js
+++ b/addon/caluma-query/queries/form.js
@@ -1,0 +1,36 @@
+import BaseQuery from "./base";
+
+export default class FormQuery extends BaseQuery {
+  dataKey = "allForms";
+  modelName = "form";
+
+  get query() {
+    return `
+      query (
+        $filter: [FormFilterSetType], 
+        $order: [FormOrderSetType], 
+        $pageSize: Int, 
+        $cursor: String
+      ) {
+        allForms(
+          filter: $filter, 
+          order: $order, 
+          first: $pageSize,
+          after: $cursor,
+        ) {
+          ${this.pagination}
+          edges {
+            node {
+              id
+              __typename
+              ...FormFragment
+            }
+          }
+        }
+      }
+
+      fragment FormFragment on Form ${this.modelClass.fragment} 
+
+    `;
+  }
+}
--- a/addon/components/cfb-form-list.js
+++ b/addon/components/cfb-form-list.js
@@ -1,28 +1,42 @@
+import { action } from "@ember/object";
 import Component from "@glimmer/component";
 import { tracked } from "@glimmer/tracking";
 import { queryManager } from "ember-apollo-client";
-import { restartableTask } from "ember-concurrency";
+import { timeout } from "ember-concurrency";
+import { restartableTask } from "ember-concurrency-decorators";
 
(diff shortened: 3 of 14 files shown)


addon/caluma-query/models/form.js
```diff
@@ -0,0 +1,11 @@
+import CalumaQueryModel from "./index";
+
+export default class FormModel extends CalumaQueryModel {
+  static FormInfo = `{
```
anehx, line 4: Please name this `fragment` to be consistent with the other models

addon/caluma-query/queries/form.js
```diff
@@ -0,0 +1,91 @@
...
+    );
+
+    // select forms whose slug contains search value and
+    // use slugs as filter for pagination query
+    forms = forms.filter((form) => form.node.slug.includes(search));
+    filter.push({ slugs: forms.map((form) => form.node.slug) });
+
+    // perform form fetch with pagination
+    return super.fetch({ filter, order, queryOptions });
+  }
```
anehx, line 90: IMO this should not be necessary. `search` can be used like every other filter provided by the API. You should be able to just call `fetch({ filter: [{ search: "xy" }] })`

addon/mirage-graphql/filters/base.js
```diff
@@ -16,6 +16,8 @@ export default class {
   }
 
   filter(records, filters) {
+    // flatten array of filters
+    filters = Array.isArray(filters) ? Object.assign(...filters) : filters;
```
anehx, line 20: Why is this needed?

addon/components/cfb-form-list.js
```diff
@@ -1,29 +1,55 @@
...
+  @action
+  loadMoreForms() {
+    this.fetchForms.perform(true);
   }
 
   @restartableTask
-  *forms() {
-    if (this.args.forms) {
-      return this.args.forms.perform();
+  *fetchForms(loadMore = false) {
```
anehx, line 39: I don't really see why this task is needed anymore. This should all be handled by the query object. Let's discuss this!

tests/integration/components/cfb-form-list-test.js
```diff
@@ -18,16 +18,16 @@ module("Integration | Component | cfb-form-list", function (hooks) {
...
-          { node: { id: 4, slug: "form-4", title: "Form 4" } },
-          { node: { id: 5, slug: "form-5", title: "Form 5" } },
+          { id: 1, slug: "form-1", title: "Form 1" },
+          { id: 2, slug: "form-2", title: "Form 2" },
+          { id: 3, slug: "form-3", title: "Form 3" },
+          { id: 4, slug: "form-4", title: "Form 4" },
+          { id: 5, slug: "form-5", title: "Form 5" },
         ];
       })
     );
```
anehx, line 28: Let's use mirage in this test.

addon/templates/components/cfb-form-list.hbs
```diff
@@ -26,30 +30,51 @@
...
-      >
-        {{uk-icon "search" ratio=10}}
-        <p class="uk-text-muted">{{t "caluma.form-builder.form.empty"}}</p>
-      </div>
-    {{/if}}
+    <div
+      data-test-form-list-empty
+      class="uk-padding-large uk-padding-remove-horizontal uk-text-center"
+    >
+      {{uk-icon "search" ratio=10}}
```
anehx, line 74:
```suggestion
      <UkIcon @icon="search" @ratio={{10}} />
```

anehx (changes requested):
Looks good in general but still needs some tweaking, we'll discuss this together and go trough my comments. Try to no overdo it with the comments those are only helpful if it's not obvious, otherwise it's just noise. E.g: `// perform form fetch with pagination` doesn't really add an value.

anehx:
Backend search will be fixed by https://github.com/projectcaluma/caluma/pull/1392

anehx (approved):
Awesome work! :100: :fire:

PR #1259, opened 2021-02-24
Fix various issues with the CI pipelines

anehx:
- f47a99ec017e311e15d450f4452fa1028170e273 optimizes build time with caching and prevents jobs from idling up to 6 hours
- b9c77481d0fed1448ad87e1ad30903b06dfaedcb fixes the idling piplines we currently have
- 69c2734fe592acf414edaab9ae135c5c636239d6 fixes the flaky test we weren't able to reproduce forever

This speeds our pipeline up from originally ~20min to now ~7min (with 2 jobs not even having a cache yet)


czosel (approved):
:100: Looks great!

PR #1260, opened 2021-02-24
chore(tests): fix flaky tests with date questions

anehx:
Date questions didn't work in our tests. Since we randomly take a type
this issue only popped up sometimes. This handles date questions
properly and also tests them!


anehx:
Superseeded by #1259

PR #1267, opened 2021-02-25
fix(date): use UTC for dates to avoid timezone problems

anehx:
This will fix our browserstack pipelines and also re-enables MS Edge testing since it now works!


czosel (approved):
:+1:

Issue #1264, opened 2021-02-25
Slug validation does not consider namespaces


anehx:
Currently, it's possible to create slugs on forms or questions etc. that are longer than they should be. This happens when ember-caluma was namespaced. If you create a slug that is e.g 2 chars less than the limit and you have a namespace of "123" it will be 2 chars longer than the actual maximum ("123-..."). Creating it will work, however, this object can't ever be edited again.

anehx:
Should be done for options as well, but with the question slug (including namespace) as prefix

PR #1283, opened 2021-03-02
fix(cfb): consider namespace during slug validation

luytena:
When building forms and questions, the namespace is taken into
account during slug validation. This ensures that the combined
length of the namespace (if existent) and slug never exceeds
the maximum length in a valid form or question.

Issue #1264

### DIFF
--- a/addon/validators/slug.js
+++ b/addon/validators/slug.js
@@ -1,15 +1,35 @@
+import buildMessage from "ember-changeset-validations/utils/validation-errors";
 import {
   validatePresence,
-  validateLength,
   validateFormat,
 } from "ember-changeset-validations/validators";
 
 import and from "ember-caluma/utils/and";
 
+const validateNamespaceSlugLength = function ({ max, namespace }) {
+  return (key, newValue, oldValue, changes, content) => {
+    const prefix =
+      (changes?.[namespace] ||
+        (changes?.[namespace] === undefined && content?.[namespace])) ??
+      "";
+
+    const slug = prefix + newValue;
+
+    return (
+      slug.length <= max ||
+      buildMessage(key, {
+        type: "tooLong",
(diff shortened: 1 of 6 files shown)


addon/validators/slug.js
```diff
@@ -1,15 +1,35 @@
...
 } from "ember-changeset-validations/validators";
 
 import and from "ember-caluma/utils/and";
 
+const validateNamespaceSlugLength = function ({ max, namespace }) {
+  return (key, newValue, oldValue, changes, content) => {
+    const prefix =
+      (changes?.[namespace] ||
+        (changes?.[namespace] === undefined && content?.[namespace])) ??
+      "";
```
anehx, line 14:
```suggestion
    const prefix = { ...content, ...changes }[namespace] || "";
```

tests/integration/components/cfb-form-editor/general-test.js
```diff
@@ -212,4 +212,41 @@ module("Integration | Component | cfb-form-editor/general", function (hooks) {
...
+    await blur("input[name=slug]");
+
+    assert.dom(".cfb-prefixed + span").doesNotExist();
+
+    await fillIn("input[name=slug]", "x".repeat(120));
+    await blur("input[name=slug]");
+
+    assert
+      .dom(".cfb-prefixed + span")
+      .hasText("Slug is too long (maximum is 119 characters)");
```
anehx, line 250: Did this work in another language?

luytena: With the `ember-intl-changeset-validations` package translations of the error messages are now possible. However, it doesn't seem to support locale switching once it's loaded and the `description` field of the error messages is not translated.

anehx:
Closed in favor of #2127

PR #1287, opened 2021-03-03
refactor(cf): use fetch API for file uploads

### LINKED ISSUE #593: Use fetch for file uploads
It's 2019 not 1997... :wink: 

https://github.com/projectcaluma/ember-caluma/blob/c424c0c06fd5707d73f618818ecbdfb74c2fc8db/addon/components/cf-field/input/file.js#L34-L54

luytena:
Replace XMLHttpRequest implementation for file
uploads with fetch API.

Resolves #593

### DIFF
--- a/addon/components/cf-field/input/file.js
+++ b/addon/components/cf-field/input/file.js
@@ -39,20 +39,14 @@ export default Component.extend({
       const data = new FormData();
       data.append("file", file);
 
-      const request = new XMLHttpRequest();
-
-      request.addEventListener("error", (event) => reject(event));
-      request.addEventListener("abort", (event) => reject(event));
-      request.addEventListener("load", (event) => {
-        if (event.target.status === 200) {
-          resolve(event);
-        } else {
-          reject(event);
-        }
-      });
-
-      request.open("PUT", url);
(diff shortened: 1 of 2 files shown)


addon/components/cf-field/input/file.js
```diff
@@ -72,6 +67,7 @@ export default Component.extend({
       }
     },
 
+    @waitFor
```
anehx, line 70: Using a test helper in a component - why?

luytena: The `await triggerEvent(...)` even though it apparently uses `settled()` internally it doesn't actually wait for the asynchronous function to terminate. From what I've read it doesn't realize that it has to wait for that particular promise and that's where the emberjs `waitFor` test helper decorator gives it that information. Is there a solution that can be used within the test that ensures that it waits for the `save` function to terminate?

luytena: Using the `waitUntil` test helper in the test seems to work as well.

anehx: Seems more reasonable

anehx (approved):
Good work!

PR #1309, opened 2021-03-16
feat(jexl): improve mapby and add stringify transform

anehx:
This extends the mapby transform to accept multiple keys as argument
which then returns an array instead of the single value. This will
result in nested arrays which is why we added a stringify transform that
converts such a collection into a string to be used in comparisons.


czosel (approved):
Nice ^_^

PR #1346, opened 2021-04-02
feat(form builder): add syntax highlighting for markdown and jexl

czosel:
Screenshot from 2021-04-02 18-29-11
Screenshot from 2021-04-02 18-29-04

### DIFF
--- a/addon/components/cfb-code-editor.js
+++ b/addon/components/cfb-code-editor.js
@@ -1,0 +1,25 @@
+import { action } from "@ember/object";
+import Component from "@glimmer/component";
+import { tracked } from "@glimmer/tracking";
+import { CodeJar } from "codejar";
+import Prism from "prismjs";
+import "prismjs/components/prism-jexl.min.js";
+import "prismjs/components/prism-markdown.min.js";
+
+export default class CfbCodeEditorComponent extends Component {
+  editor;
+  @tracked show = false;
+
+  @action
+  didInsertNode(element) {
+    const highlight = (editor) => Prism.highlightElement(editor);
+    this.editor = CodeJar(element, highlight);
+    this.editor.updateCode(this.args.value);
+    this.editor.onUpdate(this.args.update);
+  }
+
+  @action
+  willDestroyNode() {
+    this.editor.destroy();
+  }
+}
--- a/addon/components/cfb-code-editor.js
+++ b/addon/components/cfb-code-editor.js
@@ -3,8 +3,8 @@ import Component from "@glimmer/component";
 import { tracked } from "@glimmer/tracking";
 import { CodeJar } from "codejar";
 import Prism from "prismjs";
-import "prismjs/components/prism-jexl.min.js";
-import "prismjs/components/prism-markdown.min.js";
+import "prismjs/components/prism-jexl.js";
+import "prismjs/components/prism-markdown.js";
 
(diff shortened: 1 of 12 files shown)


addon/templates/components/cfb-form-editor/question.hbs
```diff
@@ -99,24 +99,23 @@
...
-      {{f.input
-        type="textarea"
-        label=(t "caluma.form-builder.question.infoText")
-        name="infoText"
-        class="uk-margin-remove-bottom"
-      }}
+      <f.input
+        @label={{t "caluma.form-builder.question.infoText"}}
+        @name="infoText"
+        @renderComponent={{component "cfb-code-editor" language="markdown" className="uk-margin-remove-bottom"}}
```
czosel, line 117: `class` and `...arguments` didn't seem to work here - not sure why.

addon/components/cfb-code-editor.hbs
```diff
@@ -0,0 +1,13 @@
+<div>
```
czosel, line 1: I had to co-locate the template because with glimmer components templates in addons are not found during testing (see https://discuss.emberjs.com/t/how-to-use-glimmer-components-in-addon/17464). Should I use a classic component instead?

velrest:
Looks good to me :+1: . Since my knowledge is limited about caluma maybe @anehx should give this a review too.

addon/components/cfb-code-editor.hbs
czosel: fixed, I forgot the template export in `app/`

addon/templates/components/cfb-code-editor.hbs
```diff
@@ -0,0 +1,13 @@
+<div>
```
anehx, line 1: Why the wrapping div?

addon/components/cfb-code-editor.js
```diff
@@ -0,0 +1,25 @@
+import { action } from "@ember/object";
+import Component from "@glimmer/component";
+import { tracked } from "@glimmer/tracking";
+import { CodeJar } from "codejar";
+import Prism from "prismjs";
+import "prismjs/components/prism-jexl.min.js";
+import "prismjs/components/prism-markdown.min.js";
```
anehx, line 7: That might better belong into `index.js` this is not the right place

app/styles/ember-caluma.scss
```diff
@@ -23,8 +23,12 @@
   }
 }
 
-.jexl-textarea textarea {
-  font-family: monospace;
+.cfb-code-editor {
+  font-family: "Source Code Pro", monospace;
```
anehx, line 27: Use the UIkit variable `$base-code-font-family`

ember-cli-build.js
```diff
@@ -19,6 +19,7 @@ module.exports = function (defaults) {
 
   app.import("node_modules/typeface-oxygen/index.css");
   app.import("node_modules/typeface-oxygen-mono/index.css");
+  app.import("node_modules/prismjs/themes/prism.css");
```
anehx, line 22: This will only work for the dummy app. This belongs into `index.js`

anehx (changes requested):

czosel:
All fixed, thanks for the review @anehx

anehx (approved):

PR #1348, opened 2021-04-06
chore(deps): upgrade momentjs

czosel:
The issue that made us downgrade moment has been fixed in
ember-moment-shim 3.8.0.
https://github.com/adopted-ember-addons/ember-pikaday/issues/343


czosel:
I needed to change the cache keys to make CI pass since there seems to be no way to clear the yarn cache. I'd be ok with merging it like this, next time we need to reset the keys we can revert them.

Yelinz (approved):

PR #1374, opened 2021-04-28
fix(form-builder): fix refreshing of page when submitting search form

anehx:


czosel:
@luytena you might want to take a look as well? :-)

luytena:
Looks good, I even learnt a few new things about JS :)

PR #1428, opened 2021-06-11
refactor(form): refactor various components to glimmer

anehx:
Refactor the following form components:

- cf-form
- cf-content
- cf-field-value
- cf-field
- cf-field/errors
- cf-field/info
- cf-field/input


anehx:
@czosel @velrest @luytena please add your review if possible

addon/components/cf-content.js
```diff
@@ -39,103 +38,79 @@ import { parseDocument } from "ember-caluma/lib/parsers";
...
-    if (this.document) this.document.destroy();
-    if (this.navigation) this.navigation.destroy();
-  },
+  @queryManager apollo;
 
   /**
    * The uuid of the document to display
    *
    * @argument {String} documentId
    */
```
czosel, line 51: Are the arguments picked up by addon-docs even without the variable declaration?

anehx: Yes

addon/components/cf-field-value.js
```diff
@@ -40,19 +38,18 @@ export default Component.extend({
...
-        },
-        "node.fileValue"
-      );
-      window.open(downloadUrl, "_blank");
-    },
-  },
-});
+  @action
+  async download(fileAnswerId) {
+    const { downloadUrl } = await this.apollo.query(
```
czosel, line 45: Why `query` instead of `watchQuery`?

czosel: Ah I see, for the download we only need to fetch once.

czosel (approved):
Nice! :+1:

PR #1435, opened 2021-06-14
feat(jexl): allow optional answer transforms with default value

anehx:
This allows the user to write reusable JEXLs using answer transforms that don't necessarily resolve to a question in a certain form context.

https://github.com/projectcaluma/caluma/pull/1536

### DIFF
--- a/addon/lib/document.js
+++ b/addon/lib/document.js
@@ -121,7 +121,9 @@ export default Base.extend({
   jexl: computed(function () {
     const documentJexl = new jexl.Jexl();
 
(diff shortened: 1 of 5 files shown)


addon/lib/document.js
```diff
@@ -171,6 +173,7 @@ export default Base.extend({
       if (this.parentDocument) return this.parentDocument.jexlContext;
 
       return {
+        null: null,
```
kaldras, line 176: As discussed, probably worthy of a comment since it is not so obvious why this is necessary.

anehx: Done

kaldras (approved):

Issue #1416, opened 2021-05-19
Form builder: direct link to sub and row form in question list


anehx:
In the question list we should have a direct link to the used form (`sub_form` or `row_form`) if the question is a table or form question.

![Screenshot from 2021-05-19 13-17-52](https://user-images.githubusercontent.com/6150577/118804173-a65a1180-b8a4-11eb-9f1d-cafede061b4e.png)

PR #1441, opened 2021-06-16
feat(cfb): link sub- & rowforms in question list

luytena:
For table and form questions add direct link to subform
or rowform in question list. Show correct path and
question list corresponding to sub- / rowform.

Resolves: #1416

image

### DIFF
--- a/addon/components/cfb-form-editor/question-list/item.hbs
+++ b/addon/components/cfb-form-editor/question-list/item.hbs
@@ -51,10 +51,21 @@
     {{/if}}
 
     <span
-      class="uk-width-expand uk-margin-small-right uk-text-small uk-text-muted uk-text-truncate"
+      class="{{if this.showFormLink "uk-width-auto" "uk-width-expand"}}
+        uk-margin-small-right uk-text-small uk-text-muted uk-text-truncate"
     >
       {{@question.label}}
     </span>
+    {{#if this.showFormLink}}
+      <span uk-icon="link" class="uk-flex-none"></span>
+      <a data-test-link-subform
+        href="#"
+        class="uk-width-expand uk-margin-small-right uk-text-small uk-text-truncate"
+        {{on "click" (fn (optional @on-click-form) (or @question.subForm @question.rowForm))}}
+      >
+        {{or @question.subForm.name @question.rowForm.name}}
+      </a>
+    {{/if}}
 
     <span class="uk-position-relative uk-width-auto">
       <UkBadge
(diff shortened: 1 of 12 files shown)


addon/components/cfb-form-editor/question-list/item.hbs
```diff
@@ -50,11 +50,27 @@
...
+       uk-margin-small-right uk-text-small uk-text-muted uk-text-truncate"
+  >
+    {{@question.label}}
+  </span>
+  {{#if (or isForm isTable)}}
+    <span uk-icon="link" class="uk-flex-none"></span>
+    <a data-test-link-subform
+      href="#"
+      class="uk-width-expand uk-margin-small-right uk-text-small uk-text-truncate"
+      {{on "click" (fn (optional @on-retrieve-form) (or @question.subForm @question.rowForm))}}
```
anehx, line 68: The action name is not really intuitive. How about `on-click-form`?

anehx (changes requested):
Minor nitpick, otherwise LGTM :fire:

addon/components/cfb-form-editor/question-list/item.hbs
luytena: Sure that works, I just updated it in all locations :+1:

anehx:
@luytena needs a rebase for the CI to work

luytena:
> @luytena needs a rebase for the CI to work

done, works now

anehx (approved):

luytena:
Tests failing because the question list is reset to reorder mode when the form is changed (at the moment also when a new question form is opened). It expects the question list to remain in the same mode after for example adding a question.

What is the expected behavior when following a link to another form while in the add / remove mode? Should the question list remain in the add / remove mode or reset to reorder mode? Originally I reset it to reorder mode to ensure that users are aware of the question list in the new form, and don't edit the question list assuming they are still editing the previous form.

anehx:
@luytena I think we should just display the link in reorder mode. When adding or deleting a question one normally doesn't want to check out the linked form ^^

luytena:
> @luytena I think we should just display the link in reorder mode. When adding or deleting a question one normally doesn't want to check out the linked form ^^

@anehx Makes sense, I've updated it accordingly :+1:

addon/components/cfb-form-editor/question-list/item.hbs
```diff
@@ -50,11 +50,33 @@
...
+{{#let
+  (and 
+    (eq @mode "reorder")
+    (or
+      (has-question-type @question "form")
+      (has-question-type @question "table")
+    )
+  )
+  as |showFormLink|
+}}
```
anehx, line 62: Let's put this in the JS file. This makes the template hard to read..

luytena: Fixed

anehx (approved):

PR #1448, opened 2021-06-25
feat: add document validity check component

Yelinz:
Not sure if ive done the polyfill correctly


package.json
```diff
@@ -100,6 +101,7 @@
     "ember-disable-prototype-extensions": "1.1.3",
     "ember-engines": "0.8.15",
     "ember-export-application-global": "2.0.1",
+    "ember-in-viewport": "^3.10.2",
```
anehx, line 104: This should probably be in `dependencies`, also it might be necessary to add this to the blueprint. You need to check whether the host app needs to have `ember-in-viewport` installed for it to work.

anehx (changes requested):
Minor nitpick, otherwise awesome work! :fire:

package.json
Yelinz: Moved to `dependencies`. Seem to work in the host app without it installing `ember-in-viewport`.

addon/components/document-validity.js
```diff
@@ -0,0 +1,40 @@
...
+   * @argument {Document} document
+   */
+
+  @tracked isValid = null;
+
+  @dropTask
+  *validate() {
+    const document = getOwner(this)
+      .factoryFor("caluma-model:document")
+      .create({ raw: parseDocument(this.args.document) });
```
anehx, line 34: According to the API docs, the user passes this down to the component. I'd welcome that.

addon/components/document-validity.hbs
```diff
@@ -0,0 +1,3 @@
+<div {{in-viewport onEnter=(perform this.validate)}}>
+    {{yield this.isValid}}
```
anehx, line 2:
```suggestion
  {{yield this.isValid}}
```

anehx (changes requested):
Apart from the other issues, there's also the issue of the missing user documentation which is mandatory for such public API components.

addon/components/document-validity.js
Yelinz: Wouldn't it be easier if the document generation was handled by caluma?

anehx: You are using the cf-content component which already does that for you. Generating documents is really expensive in terms of performance.

Yelinz: How would the host app access the document from cf-content then?

addon/components/document-validity.hbs
Yelinz: Applied and added documentation

addon/components/document-validity.js
Yelinz: Changed back to use document

Yelinz: Ah ok, document is being yielded in content.

anehx: Yes, exactly. Or you'd create a custom widget using this component which gets the document as well.

tests/dummy/app/templates/docs/validation.md
```diff
@@ -0,0 +1,23 @@
+# Validation
+
+Built in components to validate Caluma data.
+
+## Usage
+
+### DocumentValidity
+
+By passing a document to this component checks the validity of all its questions.
```
anehx, line 9: It's pretty relevant that it's only validated if this component enters the viewport. Add this and then the docs are alright

Yelinz: Added this line

tests/dummy/app/templates/docs/validation.md
```diff
@@ -0,0 +1,24 @@
...
+The document will only be validated after the component enters the viewport.
```
anehx, line 10:
```suggestion
This component validates the passed document for its validity as soon as it enters the viewport and will yield the validity status of the document.
```

anehx:
> Not sure if ive done the polyfill correctly

Well, did you test it??

Yelinz:
Yes, I did and it worked, but i've just mimicked the other pollyfill

anehx:
Alright, then it's correct I guess :wink: I'll approve now, but please make sure to **squash** the commits when merging!

anehx (approved):

PR #1463, opened 2021-07-26
fix(file upload): don't use formdata for minio upload

czosel:
It seems that minio expects the request body to be the plain file instead of a `FormData` object. 

This fixes a regression introduced in https://github.com/projectcaluma/ember-caluma/pull/1287


czosel:
Oops, my local copy was outdated. Updated version in #1464

PR #1464, opened 2021-07-26
fix(file upload): don't use formdata for minio upload

czosel:
It seems that minio expects the request body to be the plain file
instead of a FormData object.

This fixes a regression introduced in #1287.

### DIFF
--- a/addon/components/cf-field/input/file.js
+++ b/addon/components/cf-field/input/file.js
@@ -53,12 +53,9 @@ export default class CfFieldInputFileComponent extends Component {
     const { fileValue } = await this.args.onSave(file.name);
 
     try {
-      const data = new FormData();
-      data.append("file", file);
-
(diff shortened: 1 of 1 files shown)


luytena (approved):
Oh, I remember that I was unsure which one to use while implementing it. Apparently it makes a difference :sweat_smile:  
Looks good to me.

czosel:
Yeah, otherwise the upload still works, but the download looks funny (example for plain text file containing just one word, "plain"):

```
------WebKitFormBoundaryXNddgFmNpbkN1FEG
Content-Disposition: form-data; name="file"; filename="text.txt"
Content-Type: text/plain

plain

------WebKitFormBoundaryXNddgFmNpbkN1FEG--
```
Thanks for the review :-)

czosel:
cc @anehx

PR #1467, opened 2021-07-28
fix(form-builder): fix default values for table questions

anehx:


anehx:
Pipelines will succeed as soon as #1469 is merged and rebased

addon/components/cf-field/input/table.js
```diff
@@ -21,6 +21,10 @@ export default Component.extend({
   documentToEdit: null,
   documentToDelete: null,
 
+  parseDocument(raw) {
+    return parseDocument(raw);
+  },
```
czosel, line 26: Why the indirection?

anehx: Because we need to override it in the default table component

czosel: Ah of course, thanks!

tests/integration/components/cfb-form-editor/question/default/table-test.js
```diff
@@ -0,0 +1,20 @@
...
+    setupRenderingTest(hooks);
+
+    test("it renders", async function (assert) {
+      // Set any properties with this.set('myProperty', 'value');
+      // Handle any actions with this.set('myAction', function(val) { ... });
+
+      await render(hbs`<CfbFormEditor::Question::Default::Table />`);
+
+      assert.dom(this.element).hasText("");
+    });
```
czosel, line 18: Would it be possible to write a meaningful test for this?

PR #1478, opened 2021-08-05
fix(demo-app): added required "validation" route to router and added translations

StephanH90:
Currently when making a fresh copy of the git repo the documentation in the dummy repo doesn't actually show anything because the "validation" route is missing from `router.js`.

I added the route and added the required translations in the dummy app.


anehx:
Geez @Yelinz did you even test your documentation on #1448?? Thanks @StephanH90 for resolving this!

anehx (approved):

PR #1487, opened 2021-08-06
feat(document-validity): allow manual triggering of validation

anehx:


addon/components/document-validity.js
```diff
@@ -1,35 +1,44 @@
...
  * ```
  *
  * @class DocumentValidity
  * @yield {Boolean} isValid
+ * @yield {Function} validate
  */
 export default class DocumentValidity extends Component {
   /**
    * The document to be validated
    * @argument {Document} document
```
Yelinz, line 23: `@argument validateOnEnter` is missing

anehx: Fixed

Yelinz (approved):

PR #1490, opened 2021-08-10
feat(validation): switch default for validateOnEnter

anehx:
BREAKING CHANGE: The `validateOnEnter` parameter for the `DocumentValidity` component is now `false` instead of `true` since this is the default use case.


czosel (approved):
:+1:

PR #1544, opened 2021-10-01
fix(form-builder): remove deprecated usage of transitionToRoute

anehx:
https://deprecations.emberjs.com/v3.x#toc_routing-transition-methods

We now use `ember-engines-router-service` to have an engine internal router service instead of using the host router service. The `router` in the engine configuration can now be removed.

### DIFF
--- a/packages/-ember-caluma/app/app.js
+++ b/packages/-ember-caluma/app/app.js
@@ -15,7 +15,6 @@ export default class App extends Application {
         services: [
           "apollo",
           "notification",
-          "router",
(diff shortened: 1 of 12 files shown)


czosel (approved):
LGTM :+1: We should probably mention that the router service doesn't need to be injected anymore in the release notes, right?

anehx:
> LGTM +1 We should probably mention that the router service doesn't need to be injected anymore in the release notes, right?

It's in the commit description.

PR #1549, opened 2021-10-07
fix(form-builder): use raw chageset data when saving defaultAnswer

Yelinz:
This fixes an error when the value of defaultAnswer is 0


Yelinz:
superseded by #1562

PR #1614, opened 2021-11-02
feat(form): add new question type action button

anehx:
Add a new question type for action buttons that render a button for completing or skipping the work item related to the current document.

Backend implementation: https://github.com/projectcaluma/caluma/pull/1607

### DIFF
--- a/packages/form/addon/components/cf-field/input/action-button.hbs
+++ b/packages/form/addon/components/cf-field/input/action-button.hbs
@@ -1,0 +1,19 @@
+{{#unless @disabled}}
+  <DocumentValidity
+    @document={{@field.document}}
+    @validateOnEnter={{this.validateOnEnter}}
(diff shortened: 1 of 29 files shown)


packages/form/addon/components/cf-field/input/action-button.hbs
```diff
@@ -0,0 +1,19 @@
+{{#unless @disabled}}
+  <DocumentValidity
```
Yelinz, line 2: Might want to add an `onInvalid` notification?

Yelinz (approved):
lgtm

packages/form/addon/components/cf-field/input/action-button.hbs
anehx: No, the button will be disabled if it's invalid.

PR #1623, opened 2021-11-05
fix(form): safely call validate in document-validity

Yelinz:


packages/form/addon/components/document-validity.js
```diff
@@ -37,7 +37,9 @@ export default class DocumentValidity extends Component {
   @action
   async validate() {
     await Promise.all(
-      this.args.document.fields.map((field) => field.validate.perform())
+      this.args.document.fields.map(
+        async (field) => await field.validate.perform()
```
anehx, line 41: That doesn't do anything.. `field.validate.perform()` already returns a promise..

Yelinz: Getting an error from ember-concurrency that the validate is being canceled as its restartable, form this line of code. Would using enqueue instead be a solution?

anehx: Have you tried using `field.validate.linked().perform()`?

anehx (approved):
:tada: nice catch!

PR #1625, opened 2021-11-08
feat(form-builder): add delete functionality for uploaded files

aziiee:


czosel (approved):

packages/form/addon/components/cf-field/input/file.hbs
```diff
@@ -1,16 +1,29 @@
...
 
-    {{! label is in another component }}
-    {{! template-lint-disable require-input-label }}
-    <input class="uk-input" type="text" placeholder={{@placeholder}} readonly />
+      {{! label is in another component }}
+      {{! template-lint-disable require-input-label }}
+      <input
+        class="uk-input"
+        type="text"
+        placeholder={{this.placeholder}}
```
czosel, line 17: This means that the placeholder needs to be entered in the form builder, right? Would it make sense to add a generic fallback text (e.g. "Click to upload a file")? I think this would be fine for most use-cases.

czosel: Misunderstanding, that's exactly what we're doing.

czosel (approved):

PR #1643, opened 2021-11-16
fix(form): update design for input type "file"

aziiee:
This will change the design as follows:

image


anehx (changes requested):
Tests need to be updated, otherwise LGTM! Looks way better :100: :fire:

anehx (changes requested):
Hm, if I see correctly this will display absolutely nothing if disabled - that's not good..

aziiee:
> Hm, if I see correctly this will display absolutely nothing if disabled - that's not good..

@anehx Right.. fixed it. The input will now be disabled instead.

anehx (approved):

PR #1644, opened 2021-11-16
fix(testing): add missing DateTime mock

Stefano1990:
fix: adds a missing DateTime mock


anehx (approved):

anehx (changes requested):
Tests are failing.

anehx:
Closing this since we removed mocks for primitive types altogether. If no data is set on the models we shouldn't return random data.

PR #1656, opened 2021-11-23
fix(form): delete existing file input without error

derrabauke:
Uploading and deleting a file in a file input yielded an error. What happend was a deletion + another query which tried to set the current *value* to `null`.

We circumvent this bug now by a separate `delete` method on the `input::file` component, which does not perform another *save* action with `null` as *value* argument. The actual deletion of the file is triggered beforehand via `removeAnswerMutation`.

Follow up/regression from #1625

### DIFF
--- a/packages/form/addon/components/cf-field.hbs
+++ b/packages/form/addon/components/cf-field.hbs
@@ -12,6 +12,7 @@
             @disabled={{@disabled}}
             @context={{@context}}
             @onSave={{perform this.save}}
+            @onDelete={{perform this.delete}}
           />
         {{/let}}
       </div>
--- a/packages/form/addon/components/cf-field.js
+++ b/packages/form/addon/components/cf-field.js
@@ -68,4 +68,17 @@ export default class CfFieldComponent extends Component {
       // that's ok
     }
   }
+
+  /**
+   * Delete the value of the input. This will call validation but no endpoints for
+   * deleting any resources. This has to be done by the calling function.
+   *
+   * @method delete
+   */
+  @restartableTask
+  *delete() {
+    set(this.args.field.answer, "value", null);
+
+    yield this.args.field.validate.perform();
+  }
 }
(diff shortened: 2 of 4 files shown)


czosel (dismissed):
LGTM :+1: @anehx, can you take a quick look as well?

anehx (commented):
I don't get this. Why do we need to remove the answer instead of setting it to `null` like everywhere else??

derrabauke:
> I don't get this. Why do we need to remove the answer instead of setting it to `null` like everywhere else??
@anehx 
Because setting the value to `null` generates an error like this:
```JSON
"errors":[{"message":"null value in column \"name\" violates not-null constraint\nDETAIL:  [...]"}]
```
And as I understood, this was the issue. Otherwise he have to apply some fix in the backend!?

czosel:
Good point about the symmetry, sorry I missed this earlier. @winged is there a reason that saving empty values doesn't work for file questions?
Also, somewhat related, we should potentially discuss deprecating `RemoveAnswerMutation`.

winged:
> Good point about the symmetry, sorry I missed this earlier. @winged is there a reason that saving empty values doesn't work for file questions?
> Also, somewhat related, we should potentially discuss deprecating `RemoveAnswerMutation`.

I think the "proper" way would be to delete the file when the answer is saved as `NULL`. My suggestion earlier was to use the `RemoveAnswerMutation` as it already does delete the file.

The proper fix thus would be that setting the answer to `NULL` would also delete the file in MinIO. I would not deprecate the `RemoveAnswerMutation` however, as it does have it's purpose.

anehx:
> I think the "proper" way would be to delete the file when the answer is saved as `NULL`. My suggestion earlier was to use the `RemoveAnswerMutation` as it already does delete the file.
> 
> The proper fix thus would be that setting the answer to `NULL` would also delete the file in MinIO. I would not deprecate the `RemoveAnswerMutation` however, as it does have it's purpose.

I agree about your proposal of the proper way. However, I think we should deprecate `removeAnswer` since it won't be used anymore and was replaced by saving empty answers to ensure the integrity of the history models..

PR #1723, opened 2022-01-05
fix(core): override apollo service instead of serving a mixin

### LINKED ISSUE #529: Move away from mixins
We should find a better way for solving those problems than mixins, since mixins are an antipattern for the new class syntax.

https://github.com/projectcaluma/ember-caluma/blob/da9e188e16fc7e2e9a891d970734d8fad9c3edc9/addon/mixins/navigation-route.js#L5
https://github.com/projectcaluma/ember-caluma/blob/da9e188e16fc7e2e9a891d970734d8fad9c3edc9/addon/mixins/caluma-apollo-service-mixin.js#L11

This would then allow #500 to be merged.

anehx:
BREAKING CHANGE: The apollo service mixin was removed in favor of
directly overriding the service. For more information on how to migrate,
please visit the v11 migration guide.

Resolves #529

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,11 +2,6 @@
 
 module.exports = {
   extends: "@adfinis-sygroup/eslint-config/ember-addon",
-  rules: {
-    // TODO: https://github.com/projectcaluma/ember-caluma/issues/529
-    "ember/no-mixins": "warn",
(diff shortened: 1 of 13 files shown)


czosel (approved):
:+1: looks great!

PR #1740, opened 2022-01-07
feat(form): enable autoresize for textarea questions

JohnL17:


JohnL17:
@anehx I've found the [ember-autoresize-modifier](https://github.com/jelhan/ember-autoresize-modifier) addon which works similar to our implementation of the autoresize helper in the journal. Should I use this? What do you think?

anehx (commented):
Fancy m :8ball: :100:

anehx (approved):

PR #1742, opened 2022-01-10
fix(form): fix dependencies for ember-autoresize-modifier

JohnL17:


anehx (changes requested):
This needs to be added to the blueprint as well.

JohnL17:
> This needs to be added to the blueprint as well.

Is it ok like this?

anehx (approved):

PR #1743, opened 2022-01-10
fix(distribution): fix infinte loop when fetching groups with ember-data

anehx:


czosel (approved):
Followed the thread in ember-resources, nice! :100:

PR #1752, opened 2022-01-11
feat(cf-field): add optional hints for form fields

luytena:
Add option to create hints for certain question types. These
are displayed below the input field and can be used to provide
short, informative messages. Hints are available for all question
types except for form, static and action button questions.

image

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,7 +1,7 @@
 version: "3"
 services:
   db:
-    image: postgres:latest
+    image: postgres:14-alpine
     environment:
       - POSTGRES_USER=caluma
       - POSTGRES_PASSWORD=caluma
@@ -9,7 +9,7 @@ services:
       - dbdata:/var/lib/postgresql/data
 
   caluma:
-    image: ghcr.io/projectcaluma/caluma:dev
+    image: ghcr.io/projectcaluma/caluma:7.15.0
     ports:
       - "8000:8000"
     depends_on:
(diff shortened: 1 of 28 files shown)


anehx:
How about a question mark icon before the hint? Also, do I see correctly that the error and the hint have different font sizes?

luytena:
> How about a question mark icon before the hint? Also, do I see correctly that the error and the hint have different font sizes?

You are right, I fixed the font sizes. About the question mark icon, we were thinking that it might take up too much focus. This is how it would look like:

image

anehx:
Fair enough!

czosel (approved):
LGTM :+1:

anehx:
@luytena please rebase, fix the caluma container version to the latest version (not dev) and run `yarn update-caluma`. After that we can merge!

docker-compose.yaml
```diff
@@ -1,15 +1,15 @@
 version: "3"
 services:
   db:
-    image: postgres:latest
+    image: postgres:alpine
```
anehx, line 4:
```suggestion
    image: postgres:14-alpine
```

See https://github.com/projectcaluma/caluma/pull/1704/commits/29614e901c51c9a73d417a3174507b2cec62049d

luytena:
There were a number of merge conflicts, mostly in the `package.json` files. I resolved them by selecting the most recent version.

anehx:
@luytena ember-uikit 5.0.0 is more recent than 5.0.0-beta.4 :wink:

luytena:
Should be correctly rebased now.

anehx (approved):

anehx:
Awesome work, looking forward to using this! :100:

luytena:
Thanks for fixing the tests!

PR #1760, opened 2022-01-20
feat(deps): update ember-uikit to v5 beta

anehx:


packages/-ember-caluma/app/templates/docs/migration-v11.md
```diff
@@ -1,12 +1,56 @@
 # Migration from v9 / v10 to v11
 
+## General
+
+- Internet Explorer is not supported anymore
+- All `@projectcaluma/ember-*` packages now require `ember-uikit` v5+
+- The helpers `and` and `or` from `@projectcaluma/ember-core/utils/` were moved
+  to `@projectcaluma/ember-form-builder/utils/`
```
czosel, line 8: The two helpers seem pretty generic, why move them out of the core?

czosel (approved):
LGTM :+1:

packages/-ember-caluma/app/templates/docs/migration-v11.md
anehx: They're not helpers, they're utils for custom validations which are only used in the form-builder. I moved them, because they are only used in the form builder. I don't see a use case for them in other packages.

czosel: Ah, makes sense. thanks!

PR #1772, opened 2022-01-27
fix(form): change file download actions to query and network-only

Yelinz:
This fixes the graphql "Invariant Violation: 10" error which occurs,
when trying to download a file answer through the table header

### DIFF
--- a/packages/form/addon/components/cf-field-value.js
+++ b/packages/form/addon/components/cf-field-value.js
@@ -41,10 +41,16 @@ export default class CfFieldValueComponent extends Component {
   @action
   async download(id) {
     const { downloadUrl } = await this.apollo.query(
-      { query: getFileAnswerInfoQuery, variables: { id } },
+      {
+        query: getFileAnswerInfoQuery,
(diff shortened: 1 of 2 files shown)


anehx (commented):
I don't understand how this should fix an error, please elaborate. `watchQuery` in an action doesn't really make sense - why would we watch it?

Yelinz:
```
      node: invariant(options.fetchPolicy !== 'cache-and-network', 'The cache-and-network fetchPolicy does not work with client.query, because ' +
        'client.query can only return a single result. Please use client.watchQuery ' +
        'to receive multiple results from the cache and the network, or consider ' +
        'using a different fetchPolicy, such as cache-first or network-only.')
  },

```
This is the error which is thrown, the number changed with the apollo release.
The Problem as stated, either using a different network policy for files or using watchQuery is needed.
I have just mimicked the download action in `packages/form/addon/components/cf-field/input/file.js`, which also uses `watchQuery` in an action.

anehx (approved):

PR #1775, opened 2022-02-01
fix: dye file upload button in default color

derrabauke:


anehx (changes requested):
This is a breaking change. Did you check with all projects using this if that's ok? Maybe this should be configurable?

czosel:
We checked with all projects, all good.

PR #1778, opened 2022-02-02
fix: render static questions

czosel:


czosel:
:+1: fixed

anehx:
> +1 fixed

I'd be disappointed if the tests don't fail - but let's see..

anehx (approved):
:heart: :100:

PR #1779, opened 2022-02-03
fix(validation): sync format validator validation with backend

anehx:
BREAKING CHANGE: Use the `formatValidators` property of the backend to store and read format validators instead of the `meta.formatValidators` so the backend validates as well. For more information on how to migrate check the migration guide to v11.

Related to https://github.com/projectcaluma/caluma/pull/1665


czosel (approved):
:+1:

PR #1791, opened 2022-02-09
feat(distribution): remove distribution top level route

anehx:


luytena (approved):
Looks good to me :ok_hand:

PR #1797, opened 2022-02-15
chore(tests): add coverage plugin for distribution

anehx:


anehx:
offline approved by @czosel

PR #1798, opened 2022-02-15
feat(distribution): add config variable to stack the UI

anehx:


anehx:
offline approved by @czosel

PR #1799, opened 2022-02-15
fix(distribution): show controls if no inquiries exist yet

anehx:


anehx:
offline approved by @czosel

PR #1800, opened 2022-02-15
feat(distribution): add index page when no inquiries exist

anehx:


anehx:
offline approved by @luytena

PR #1801, opened 2022-02-15
feat(distribution): add proper 404 handling

anehx:
Depends on #1800


anehx:
offline approved by @luytena

PR #1827, opened 2022-02-24
fix(workflow): return case status on complete / skip work-item mutation

luytena:


packages/form/tests/integration/components/cf-field/input/action-button-test.js
```diff
@@ -16,9 +16,14 @@ module(
...
 
+      const workItem = this.server.create("work-item");
+      workItem.update({
+        case: this.server.create("case"),
+      });
+
       this.field = {
         document: {
-          workItemUuid: this.server.create("work-item").id,
+          workItemUuid: workItem.id,
```
anehx, line 26:
```suggestion
          workItemUuid: this.server.create("work-item", { case: this.server.create("case") }).id,
```

luytena: Done :+1:

anehx (approved):

PR #1831, opened 2022-02-25
chore: use embroider macro for skipping timeouts in tests

anehx:


packages/distribution/addon/components/inquiry-new-form.js
```diff
@@ -63,10 +63,10 @@ export default class InquiryNewFormComponent extends Component {
   *updateSearch(e) {
     e.preventDefault();
 
-    const { environment } =
-      getOwner(this).resolveRegistration("config:environment");
-
-    if (environment !== "test") {
+    /* istanbul ignore next */
```
luytena, line 66: What does `/* istanbul ignore next */` do?

packages/-ember-caluma/package.json
```diff
@@ -14,6 +14,7 @@
   "devDependencies": {
     "@ember/jquery": "2.0.0",
     "@ember/optional-features": "2.0.0",
+    "@embroider/macros": "1.2.0",
```
luytena, line 17: Why is the `^` not used in this `package.json` compared to the ones in the other packages?

anehx: This is a dev dependency, not a dependency

packages/distribution/addon/components/inquiry-new-form.js
anehx: Ignores the next block for code coverage

luytena (approved):

PR #1832, opened 2022-02-25
feat(form-builder): add new categories published and unpublished

anehx:
Add new category filters "published" and "unpublished" to the form list and make sure the filters and the search are stored in a query param for direct links.


luytena (approved):
Looks good to me!

PR #1833, opened 2022-02-25
feat(form): support passing component override classes

anehx:
This is needed so widget overrides can be used in an engine since the engine can't resolve the component by name.

### DIFF
--- a/packages/form/addon/components/cf-field.hbs
+++ b/packages/form/addon/components/cf-field.hbs
@@ -14,7 +14,10 @@
 
     <div class="uk-flex">
       <div class="uk-width-expand">
-        {{#let (component (get-widget @field.question)) as |FieldComponent|}}
+        {{#let
+          (component (ensure-safe-component (get-widget @field.question)))
+          as |FieldComponent|
+        }}
           <FieldComponent
             @field={{@field}}
             @disabled={{@disabled}}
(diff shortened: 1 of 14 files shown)


packages/form/addon/components/cf-field.hbs
```diff
@@ -14,7 +14,10 @@
 
     <div class="uk-flex">
       <div class="uk-width-expand">
-        {{#let (component (get-widget @field.question)) as |FieldComponent|}}
+        {{#let
+          (component (ensure-safe-component (get-widget @field.question)))
```
derrabauke, line 18: Is `ensure-safe-component` still needed, since you already cover it in the `get-widget` helper?

packages/form/addon/components/cf-form-wrapper.hbs
```diff
@@ -1,6 +1,8 @@
 {{#let
   (component
-    (get-widget @fieldset.field.question @fieldset.form default="cf-form")
+    (ensure-safe-component
```
derrabauke, line 3: Same as above (`ensure-safe-component` needed?)

derrabauke (approved):
Impressive job! :+1: 
Just a question of understanding regarding the double use of `ensure-safe-component` (see comments).

packages/form/addon/components/cf-field.hbs
anehx: Yes, because we use it in a `component` helper.

packages/form/addon/components/cf-form-wrapper.hbs
anehx: Same as above

packages/form/addon/components/cf-field.hbs
derrabauke: Alright. Thanks for clarification. :+1:

PR #1835, opened 2022-02-25
fix(core): handle null values in raw dates

### LINKED ISSUE #1826: add null check for dates from raw data
https://github.com/projectcaluma/ember-caluma/blob/d7d5c665ba6406f57fa0271657dce108fa82cf07/packages/core/addon/caluma-query/models/index.js#L17

there needs to be a check if `this.raw[name] !== null`, otherwise is will create a date object with ` new Date(null).valueOf() === 0`

anehx:
Resolves #1826


derrabauke (approved):
:rocket: Thanks for the fix!

PR #1879, opened 2022-03-24
fix(distribution): make sure distribution components are prefixed

anehx:
This makes sure that our component names don't collide with the hosts components


luytena (commented):
Is name collision for components always an issue for ember-engines?

anehx:
> Cd = Caluma Distribution?

Exactly, like `cf-` in `caluma-form` and `cfb-` in `caluma-form-builder`

anehx:
> Is name collision for components always an issue for ember-engines?

It's always an issue for all kinds of addons, regardless if it's a normal addon or an engine. The component trees will always be merged with the host app which causes a collision if anything is named the same.

luytena:
> > Is name collision for components always an issue for ember-engines?
> 
> It's always an issue for all kinds of addons, regardless if it's a normal addon or an engine. The component trees will always be merged with the host app which causes a collision if anything is named the same.

Oh, good to know :sweat_smile:

PR #1880, opened 2022-03-25
feat(form): show technical details of backend errors in UI

anehx:
error

### DIFF
--- a/packages/form/addon/components/cf-field.hbs
+++ b/packages/form/addon/components/cf-field.hbs
@@ -38,7 +38,25 @@
           {{#if this.save.isRunning}}
             <UkSpinner class="uk-animation-fade" />
           {{else if (or this.save.last.isError @field.isInvalid)}}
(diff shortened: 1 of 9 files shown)


anehx:
BTW, for those who see this kind of stuff: the slight offset between the info and error icon was fixed after the demo was recorded :wink:

czosel (approved):
Nice!

czosel:
Only idea, now that I also checked out the video: Should we add a message like "The answer could not be saved" before "Technical details"? Otherwise, people might be confused about what the "!" icon means exactly.

anehx:
> Only idea, now that I also checked out the video: Should we add a message like "The answer could not be saved" before "Technical details"? Otherwise, people might be confused about what the "!" icon means exactly.

Good point. I added a message for the user and updated the GIF

czosel (approved):

PR #1881, opened 2022-03-25
fix(testing): mock DateTime scalar

Yelinz:


anehx (changes requested):
This will result in every datetime being the current time - we don't want this. Why do you need this?

Yelinz:
Noticed that as well, still working on this. 
The problem is that when using this addon it will throw the error `No mock defined for type "DateTime"`
which led me to monkey patching a DateTime mock there and that resolved the error

Yelinz:
Date and DateTime aren't default graphql scalar types and DateTiime isn't defined in `mirage-graphql/mocks/base.js` which lead me to belive it wasn't being mocked and added it here

anehx:
That means that there is no data on the model. What field on what model did you try to get?

Yelinz:
case createdAt and modifiedAt, which should be generated by the factory when I generate them.

Yelinz:
Overlooked closedAt, closing as its not a bug. Started some updating of graphql-tools, will commit that in a other PR.

PR #1892, opened 2022-03-31
fix(form-builder): replace prismjs with highlight.js

### LINKED ISSUE #1864: Documentation code snippet visibility problem
When using Firefox with the default dark theme provided by Mozilla the code snippets in the documentation becomes unreadable.
Firefox version 98.0
Mozilla dark theme version 1.2
image

Yelinz:
This is to prevent them from clashing in the addon docs, as they use the
same tags.

How the form-builder editors with highlighting look like now
image
image

fixes #1864

### DIFF
--- a/packages/form-builder/addon/components/cfb-code-editor.js
+++ b/packages/form-builder/addon/components/cfb-code-editor.js
@@ -2,11 +2,14 @@ import { action } from "@ember/object";
 import { addObserver } from "@ember/object/observers";
 import Component from "@glimmer/component";
 import { CodeJar } from "codejar";
-import Prism from "prismjs";
+import hljs from "highlight.js/lib/core";
+import json from "highlight.js/lib/languages/json";
+import markdown from "highlight.js/lib/languages/markdown";
+import jexl from "highlightjs-jexl/src/languages/jexl";
 
-import "prismjs/components/prism-json.js";
-import "prismjs/components/prism-jexl.js";
-import "prismjs/components/prism-markdown.js";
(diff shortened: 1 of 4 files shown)


Yelinz:
Documentation looks normal now aswell
image

packages/form-builder/addon/components/cfb-code-editor.js
```diff
@@ -2,11 +2,14 @@ import { action } from "@ember/object";
 import { addObserver } from "@ember/object/observers";
 import Component from "@glimmer/component";
 import { CodeJar } from "codejar";
-import Prism from "prismjs";
+import hljs from "highlight.js/lib/core";
+import json from "highlight.js/lib/languages/json";
+import markdown from "highlight.js/lib/languages/markdown";
+import jexl from "highlightjs-jexl/src/languages/jexl";
```
anehx, line 8: Can we try to get this into `hightlighjs`?

czosel: Highlight doesn’t accept new languages in its core anymore. One can make PRs that list a new language as supported via a plugin though - I didn’t do that yet because our plugin is still lacking tests. Would love to do that though, but I might not find the time.

czosel (approved):
Nice! 💯

packages/form-builder/addon/components/cfb-code-editor.js
anehx: Alright, I didn't know that.

PR #1946, opened 2022-05-02
feat(distribution): enable reminders for overdue inquiries

luytena:
Add a configurable functionality for sending reminders on overdue inquiries. A list of previous reminders are displayed and need to be updated through the meta field on the inquiry.
    
Co-authored-by: Jonas Metzener <jonas.metzener@adfinis.com>


anehx (approved):
:100: :fire:

PR #1948, opened 2022-05-02
feat(distribution): add button to create new inquiry in dialog

anehx:
This adds a button to create a new inquiry for the target group of the current dialog directly in the dialog itself.

### DIFF
--- a/packages/distribution/addon/components/cd-inquiry-dialog.js
+++ b/packages/distribution/addon/components/cd-inquiry-dialog.js
@@ -4,15 +4,24 @@ import { queryManager, getObservable } from "ember-apollo-client";
 import { dropTask } from "ember-concurrency";
 import { useTask } from "ember-resources";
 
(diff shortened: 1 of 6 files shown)


packages/distribution/addon/components/cd-inquiry-dialog.js
```diff
@@ -62,4 +72,47 @@ export default class CdInquiryDialogComponent extends Component {
...
+        },
+        decodeId(this.inquiries[0].id)
+      );
+    } catch (e) {
+      this.notification.danger(
+        this.intl.t("caluma.distribution.new.error", {
+          count: 1,
+        })
+      );
+    }
```
luytena, line 116: The code is very similar to the submit task of the cd-inquiry-new-form. Can we somehow centralize the create inquiry functionality and reuse it?

anehx: Fair - done!

PR #1977, opened 2022-05-19
fix(form): improve validation of choice and multiple choice fields

anehx:
- Display the label of the faulty option in the error message
- Don't accept disabled options as valid values in validation
- Disable inputs of disabled options

### DIFF
--- a/packages/form/addon/lib/field.js
+++ b/packages/form/addon/lib/field.js
@@ -582,10 +582,10 @@ export default class Field extends Base {
   @cached
   get errors() {
     return this._errors.map(({ type, context, value }) => {
-      return this.intl.t(
-        `caluma.form.validation.${type}`,
-        Object.assign({}, context, { value })
-      );
+      return this.intl.t(`caluma.form.validation.${type}`, {
+        ...context,
+        value,
+      });
     });
   }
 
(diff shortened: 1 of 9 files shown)


packages/form/addon/lib/field.js
```diff
@@ -801,10 +790,20 @@ export default class Field extends Base {
...
+    const label = Array.isArray(this.selected)
+      ? this.selected.find((selected) => selected.slug === value)?.label
+      : this.selected?.label;
+
+    return validate("inclusion", value, {
+      in: (this.options || [])
+        .filter((option) => !option.disabled)
+        .map(({ slug }) => slug),
+      allowBlank,
+      label: label ?? value,
```
luytena, line 806: For what is the `label` option in the `validate` function?

anehx: It's used in the translation

packages/form/translations/fr.yaml
```diff
@@ -46,7 +46,7 @@ caluma:
       greaterThanOrEqualTo: "La valeur indiquée dans ce champ ne doit pas être inférieure à {gte} "
       lessThanOrEqualTo: "L'entrée dans ce champ ne doit pas être supérieure à {lte} "
       notAnInteger: "Veuillez entrer un nombre entier"
-      inclusion: "'{value}' n'est pas une valeur valide pour ce champ"
+      inclusion: '"{label}" n''est pas une valeur valide pour ce champ'
```
luytena, line 49: Double `'` so that the `n'est` is displayed correctly?

anehx: Yes, `'` is the escape character in translations: https://formatjs.io/docs/core-concepts/icu-syntax/#quoting--escaping

luytena (approved):

PR #2000, opened 2022-05-31
feat(distribution): add configurable status for unanswered inquiries

anehx:
distribution-status


packages/distribution/addon/components/cd-inquiry-dialog/inquiry-part.hbs
```diff
@@ -1,8 +1,14 @@
 <div class="uk-margin-remove-last-child">
-  <p class="uk-flex uk-flex-middle uk-text-large uk-margin-remove">
+  <p
+    class="uk-flex uk-flex-middle uk-text-large uk-margin-remove"
+    data-test-title
+  >
```
luytena, line 5: Can we please have the same distance between the label and the title in the inquiry dialog as in the detail view?

luytena:
Please mirror status in inquiry dialog and in detail view.

PR #2040, opened 2022-07-05
(feat): multi file question type

derrabauke:
This is the frontend part to [this caluma PR](https://github.com/projectcaluma/caluma/pull/1821) which changes the `file-question` type to enable multiple file uploads. This leads to a series of changes around the graphql setup and the whole file handling procedure.

BREAKING CHANGE: This renames the question type constant for file questions,
and changes the semantics of the answer value for file questions as well:
It is now an array of objects instead of a string. The response type
for querying file(s) answers now also is a list.

For details / RFC see https://github.com/projectcaluma/caluma/issues/1780

Design Preview:
fileupload_1

fileupload_2


derrabauke:
[Parallel caluma PR](https://github.com/projectcaluma/caluma/pull/1809/files)

packages/form/addon/components/cf-field-value.js
```diff
@@ -2,22 +2,23 @@ import { action } from "@ember/object";
...
 
   @action
   async download(id) {
-    const { downloadUrl } = await this.apollo.query(
+    const files = await this.apollo.query(
       {
-        query: getFileAnswerInfoQuery,
-        variables: { id },
+        query: getFilesAnswerInfoQuery,
+        variables: { id: this.args.field.answer.raw.id },
```
czosel, line 15: why not `id` as before?

czosel: nevermind, I should've read a little further

packages/form/addon/components/cf-field/input/file.js
```diff
@@ -4,64 +4,92 @@ import Component from "@glimmer/component";
...
-    return this.args.field?.answer?.value?.downloadUrl;
+  get files() {
+    return this.args.field?.answer?.value;
   }
 
-  get downloadName() {
-    return this.args.field?.answer?.value?.name;
+  get allowedMimeTypes() {
+    // TODO: put this into a config env value
+    return ["image/png", "image/jpeg", "text/plain", "application/pdf"];
```
czosel, line 20: This looks unfinished :-) Also, is there a reason we're "sneaking in" this new  feature here?

packages/form/addon/components/cf-field/input/file.js
```diff
@@ -4,64 +4,92 @@ import Component from "@glimmer/component";
...
     if (downloadUrl) {
       window.open(downloadUrl, "_blank");
     }
   }
 
   @action
   async save({ target }) {
-    const file = target.files[0];
+    // store the old list of files
+    const fileList = this.files?.length ? this.files.slice() : [];
```
czosel, line 46: I'd use destructuring for copying, something like
```suggestion
    const fileList = [...this.files] || []
```

derrabauke: `this.files` can be `undefined` in some circumstances. As it isn't a `Iterable` then, it can not be destructured. So we will still need the `this.files?.length` check.

czosel: Right, how about this then?
```js
const fileList = [...this.files || []] 
```

derrabauke: Well this looks pretty!

packages/form/addon/components/cf-field/input/file.js
```diff
@@ -4,64 +4,87 @@ import Component from "@glimmer/component";
...
-      this.args.field.answer.value = {
-        name: file.name,
-        downloadUrl: fileValue.downloadUrl,
-      };
+        // replace fileList value with filesValue value
+        fileList.splice(
+          fileList.findIndex((f) => f.name === file.name),
+          1,
+          file
+        );
```
anehx, line 83: What exactly is this doing?

anehx (changes requested):
Needs some fixes for consistency and readability but otherwise LGTM. Could you add a screenshot (especially of the new file list). Please also change the commit messages - currently this won't even trigger a release..

packages/form/addon/components/cf-field/input/file.js
```diff
@@ -4,64 +4,87 @@ import Component from "@glimmer/component";
...
+    // unwrap files from FileList construct
+    const newFiles = [];
+    for (let i = 0; i < target.files.length; i++) {
+      newFiles.push({ name: target.files[i].name });
+      if (
+        !fileList.find((existing) => existing.name === target.files[i].name)
+      ) {
+        fileList.push({ name: target.files[i].name });
+      }
+    }
```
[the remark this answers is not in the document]
derrabauke: Totally valid! Refactored it.

packages/form/addon/components/cf-field/input/file.js
derrabauke: This was my unnecessary approach to combine the graph-QL answer with the file list from the input. But actually we should set the input value to the actual graph-QL answer value. `Code deleted`

packages/form/addon/components/cf-field/input/files.js
```diff
@@ -41,48 +41,47 @@ export default class CfFieldInputFilesComponent extends Component {
...
+    }));
+
     if (!newFiles) {
       return;
     }
 
-    // trigger save action for deduped file list of old and new files reduces properties
+    newFiles
+      .filter(
+        (file) => !fileList.find((existing) => existing.name === file.name)
```
czosel, line 55: Consider using presence of `id` attribute to figure out if a file is new or not

derrabauke: As there is no need to restrict the behavior in terms of duplicated file upload, I will leave out the filtering per se. As this can lead to duplicated names in the file list I have to refactor the component a little bit for deleting and downloading (`file.name` was used as identifier).

packages/form/addon/components/cf-field/input/files.js
```diff
@@ -0,0 +1,111 @@
...
+    );
+
+    try {
+      // iterate over list of new files and map to graphql answer values
+      // this way we only upload new files selected by the input
+      const _newFiles = newFiles.map((file) => ({
+        ...filesValue.find(
+          (value) =>
+            file.name === value.name &&
+            fileList.findIndex((file) => file.id === value.id) === -1
```
czosel, line 73: maybe this is simpler?
```suggestion
            !fileList.find((file) => file.id === value.id)
```

packages/form/addon/components/cf-field/input/files.js
```diff
@@ -28,8 +29,13 @@ export default class CfFieldInputFilesComponent extends Component {
       },
       "node.value"
     );
-
-    const { downloadUrl } = answers.find((file) => file.name === fileName);
+    const { downloadUrl } =
+      answers.find((file) =>
+        macroCondition(isTesting())
+          ? file.id === fileId ||
+            atob(file.id).substring(file.__typename.length + 1) === fileId
```
czosel, line 36: Why is this needed?

packages/form/addon/components/cf-field/input/files.js
```diff
@@ -0,0 +1,111 @@
...
+  async save({ target }) {
+    // store the old list of files
+    const fileList = [...(this.files || [])];
+
+    // unwrap files from FileList construct
+    const newFiles = Array.from(target.files).map((file) => ({
+      name: file.name,
+      value: file,
+    }));
+    newFiles.forEach((file) => fileList.push({ name: file.name }));
```
czosel, line 54:
```suggestion
   fileList = [...fileList, ...newFiles]
```
or if you really can't have the `id` property in there
```suggestion
   fileList = [...fileList, ...newFiles.map(({ name }) => ({ name })]
```

packages/form/addon/components/cf-field/input/files.js
```diff
@@ -0,0 +1,111 @@
...
+
+    if (newFiles.length === 0) {
+      return;
+    }
+
+    // trigger save action for file list of old and new files with
+    // reduces properties to match gql format
+    const { filesValue } = await this.args.onSave(
+      fileList.map(({ name, id }) => ({ name, id }))
```
czosel, line 63: as far as i can tell, the only property you're filtering out here is `value`. Is that really causing problems in the callsite?

packages/form/addon/components/cf-field/input/files.js
```diff
@@ -0,0 +1,111 @@
...
+      }
+      this.args.field.answer.value = filesValue;
+    } catch (error) {
+      await this.args.onSave([]);
+      this.args.field._errors = [{ type: "uploadFailed" }];
+    } finally {
+      // eslint-disable-next-line require-atomic-updates
+      target.value = "";
+    }
+  }
```
czosel, line 97: I find this method in general a bit unclear. We have `this.files`, `fileList`, `newFiles` and `_newFiles`. Maybe we can come up with more clear / speaking naming? Or maybe we don't need that many variables?

packages/form/addon/components/cf-field/input/files.js
derrabauke: Cause the testing graph-ql mock does a base64 encoding of the `{__typename: file.id}` which the real backend does not. I didn't want to mess with the testing setup though.

derrabauke: I added a comment in the code

packages/form/addon/components/cf-field/input/files.js
derrabauke: It does indeed has to have this certain form.

We defined the answer input like this in `packages/testing/addon/mirage-graphql/schema.graphql`:
```gql
input SaveDocumentFilesAnswerInput {
  question: ID!
  document: ID!
  meta: JSONString
  value: [SaveFile]
  clientMutationId: String
}
input SaveFile {
  id: String
  name: String
}
```

derrabauke: image

Indeed it needs this certain format since we defined it in `packages/testing/addon/mirage-graphql/schema.graphql`:
```gql
input SaveDocumentFilesAnswerInput {
  question: ID!
  document: ID!
  meta: JSONString
  value: [SaveFile]
  clientMutationId: String
}
input SaveFile {
  id: String
  name: String
}
```

packages/form/addon/components/cf-field/input/files.js
derrabauke: I totally aggree! Tried to rename some variables and refactor to reuse as much as possible.

packages/form/addon/components/cf-field/input/files.js
derrabauke: First option works fine! :+1:

czosel (approved):
:100: Looks good!

PR #2052, opened 2022-07-21
fix(form): include disabled options in validation

luytena:
Include disabled options in validation of checked multiple choice
and choice questions. This allows a successful validation, even
when an option is disabled, after the option has been previously
selected (and therefore can no longer be changed).

image


anehx:
I don't think this is the correct approach. If the option is disabled at the moment we validate, it can't be accepted. However, the user should be able to uncheck the disabled option, this needs to be fixed

PR #2068, opened 2022-08-03
fix(deps): resolve dep issue of ember-cli-addon-docs and tracked-toolbox

derrabauke:
## Description

After latest dependency updates the CI went ok, but running the local docs app could not be accessed anymore. Until further investigation into the issue we downgrade the dependency.

Fixes # `Uncaught TypeError: decorator is not a function`

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,10 +56,12 @@
     "semver": "7.3.7"
   },
   "//": [
-    "TODO: remove validated-changeset resolution when https://github.com/poteto/ember-changeset/issues/650 is addressed"
+    "TODO: remove validated-changeset resolution when https://github.com/poteto/ember-changeset/issues/650 is addressed",
+    "TODO: remove tracked-toolbox resolution when https://github.com/ember-learn/ember-cli-addon-docs/issues/1309 is addressed"
   ],
   "resolutions": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -59,7 +59,8 @@
     "TODO: remove validated-changeset resolution when https://github.com/poteto/ember-changeset/issues/650 is addressed"
   ],
   "resolutions": {
-    "**/validated-changeset": "~1.3.2"
+    "**/validated-changeset": "~1.3.2",
+    "**/highlight.js": "10.7.2"
```
czosel, line 63: Can you add a comment explaining the downgrade similar to how it’s done for validated-changeset? A link to a GitHub issue would be ideal. If no issue exists until now, maybe open one?

derrabauke:
Figured out, it must be an issue with tracked-toolbox and ember-cli-addon-docs:
- opened [an issue](https://github.com/ember-learn/ember-cli-addon-docs/issues/1309)
- added resolution for `tracked-toolbox`

package.json
derrabauke: Done

czosel (approved):

PR #2083, opened 2022-08-11
feat(testing): allow inverting of filters

StephanH90:
## Description

Implements the `invert` option of the real caluma backend.

## Depedencies

No dependencies.


anehx (changes requested):
Can you remove the existing implemenations of inversion? Otherwise those don't invert at all..

StephanH90:
fixed the filters and added required tests.

anehx (approved):
Nice! :100:

PR #2086, opened 2022-08-12
fix(files): disable delete button when field is disabled

### LINKED ISSUE #2084: Multiple-Files field: Deletion still possible when disabled
When the field is disabled, uploading new files becomes impossible, as expected. However, the "delete" button for each file remain active, which may lead to unexpected data loss. (Obviously the backend should reject the call, but it shouldn't even be possible to click).

derrabauke:
## Description

When a field is disabled one could still trigger the delete file action.

Fixes #2084

### DIFF
--- a/packages/form/addon/components/cf-field/input/files.hbs
+++ b/packages/form/addon/components/cf-field/input/files.hbs
@@ -23,12 +23,15 @@
         >
           {{file.name}}
         </UkButton>
-        <UkIcon
-          class="uk-icon-button uk-margin-small-left"
-          role="button"
-          @icon="trash"
-          {{on "click" (fn this.delete file.id)}}
-        />
+        {{#unless @disabled}}
+          <UkIcon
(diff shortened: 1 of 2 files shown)


packages/form/addon/components/cf-field/input/files.hbs
```diff
@@ -24,7 +24,8 @@
           {{file.name}}
         </UkButton>
         <UkIcon
-          class="uk-icon-button uk-margin-small-left"
+          class="uk-icon-button uk-margin-small-left
+            {{if @disabeld 'uk-disabled'}}"
```
winged, line 28: Could we just hide those buttons altogether if disabled?

derrabauke: Sure. Let's do that.

winged (approved):
LGTM :+1:

PR #2097, opened 2022-08-26
fix(analytics): improve alias save

Yelinz:
## Description
use different icon as a checkmark is confusing

image
this looks like it has been auto saved, just like in the form
image
looks more something you have to click than before

as uikit doesnt have a classical save icon, which would be better, this just uses something similar
I'm open to suggestions to use a different icon

### DIFF
--- a/packages/analytics/addon/components/ca-field-selector-list.hbs
+++ b/packages/analytics/addon/components/ca-field-selector-list.hbs
@@ -21,7 +21,7 @@
           <td>
             <CaFieldSelectorList::CaFieldAliasInput
               @value={{field.alias}}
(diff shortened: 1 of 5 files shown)


anehx:
Have you tried `download`. I think this one will confuse as well.. ^^

anehx:
I just discussed this with @kaldras, why aren't we autosaving? Also, a button in an input field is somehow not really intuitive, why don't we just put a save button next to it?

Yelinz:
Not sure, changed to auto save, to be consistent with the other fields

anehx (approved):
LGTM :+1:

PR #2102, opened 2022-08-30
feat(distribution): add functionality to reopen an inquiry

anehx:


packages/distribution/addon/components/cd-inquiry-dialog.js
```diff
@@ -68,11 +70,42 @@ export default class CdInquiryDialogComponent extends Component {
...
+       * now not anymore. This indicates that the work item was redone which
+       * should result in a transition to the answer view.
+       */
+      const redoableChangedToTrue = this._redoableStates.find((stateObj) => {
+        const updatedWorkItem = allWorkItems.edges.find(
+          (edge) => decodeId(edge.node.id) === stateObj.id
+        );
+
+        return (
+          stateObj.isRedoable && !(updatedWorkItem?.node.isRedoable ?? true)
```
luytena, line 85: Which case is handled by the `?? true`? Is the `isRedoable` property reset in certain work-items?

anehx: If `updatedWorkItem` is `undefined`

luytena: Ah yes, makes sense, thanks.

luytena: Minor nitpick: Shouldn't the name be `redoableChangedToFalse` or `redoneWorkItem`, if we're looking for work-items that are no longer redoable and have been redone?

luytena:
Otherwise looks good to me :ok_hand:

packages/distribution/addon/components/cd-inquiry-dialog.js
anehx: Valid point, changed it to `redoneWorkItem`

luytena (approved):

PR #2131, opened 2022-09-23
Minor distribution improvements

anehx:


packages/distribution/addon/components/cd-navigation/section.js
```diff
@@ -1,10 +1,42 @@
...
+
 export default class CdNavigationSectionComponent extends Component {
+  @service router;
+
   @tracked expanded = true;
 
+  get isActive() {
+    try {
+      const { from = null, to = null } = findInquiryParams(
+        this.router.externalRouter.recognize(this.router.currentURL)
```
luytena, line 25: For better understanding: Why is this approach needed instead of using currentRoute?

anehx: Good question.. I found an even simpler way to solve this. I now used the `isActive` method of the router to check the active state of the sections inquiries.

luytena (approved):
LGTM :ok_hand:

PR #2134, opened 2022-09-29
feat(form-builder): add validation for JEXL fields

anehx:


anehx:
Offline-approved by @czosel

PR #2141, opened 2022-10-06
feat(form): replace unmaintained pikaday datepicker with flatpickr

anehx:
**Before:**

Screenshot from 2022-10-06 17-09-16

**After:**

Screenshot from 2022-10-06 17-25-00


anehx:
Offline-approved by @czosel

PR #2156, opened 2022-10-17
feat(distribution): improve navigation

anehx:
- Change translations for sections
- Hide sections without content
- Hide navigation on index page if no inquiries exist


packages/distribution/addon/controllers/index.js
```diff
@@ -0,0 +1,6 @@
+import Controller from "@ember/controller";
+import { inject as service } from "@ember/service";
+
+export default class IndexController extends Controller {
+  @service distribution;
+}
```
luytena, line 6: No longer needed

anehx: Removed

PR #2157, opened 2022-10-17
feat(distribution): add button to mark inquiries as checked

anehx:
Depends on #2156

### DIFF
--- a/packages/distribution/addon/abilities/distribution.js
+++ b/packages/distribution/addon/abilities/distribution.js
@@ -30,6 +30,15 @@ export default class DistributionAbility extends Ability {
     );
   }
 
+  get canCheckInquiries() {
+    return (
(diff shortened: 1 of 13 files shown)


packages/distribution/app/components/cd-navigation/controls/icon.js
```diff
@@ -0,0 +1 @@
+export { default } from "@projectcaluma/ember-distribution/components/cd-navigation/controls/icon";
```
luytena, line 1: Can be removed

luytena (changes requested):

packages/distribution/app/components/cd-navigation/controls/icon.js
anehx: Removed

PR #2162, opened 2022-10-20
Distribution fixes

anehx:


packages/distribution/addon/routes/index.js
```diff
@@ -1,41 +1,28 @@
...
 export default class IndexRoute extends Route {
+  @service("-scheduler") scheduler;
+  @service distribution;
   @service router;
-  @service calumaOptions;
 
-  @config config;
+  async redirect() {
+    // trigger resource
+    this.distribution.navigation.value;
```
luytena, line 11: Why do we have to trigger the resource in this way?

anehx: How else?

luytena (approved):
:100: :fire:

PR #2167, opened 2022-10-21
feat(distribution): include inquiry answer details

luytena:


anehx (approved):
Awesome work! :100: :fire: :gun:

PR #2171, opened 2022-10-26
fix(distribution): refetch inquiry after editing for dialog preview

luytena:


packages/distribution/addon/gql/queries/inquiry-edit.graphql
```diff
@@ -12,6 +14,7 @@ query InquiryEdit($inquiry: ID!) {
         }
         document {
           id
+          ...InquiryRequest
```
anehx, line 17: Why do we need the request in this query?

packages/distribution/addon/components/cd-inquiry-edit-form.js
```diff
@@ -51,4 +55,12 @@ export default class CdInquiryEditFormComponent extends Component {
       );
     }
   }
+
+  async willDestroy() {
+    super.willDestroy();
+
+    await this.apollo.apolloClient.refetchQueries({
+      include: [inquiryEditQuery],
```
anehx, line 63: Shouldn't we refetch the dialog query?

anehx (changes requested):

packages/distribution/addon/components/cd-inquiry-edit-form.js
luytena: That works too, but then I refetch all of the inquiries on the dialog, when I only want the one inquiry. In addition, the inquiryDialogQuery seems to usually have multiple active observable queries, which then refetches all of them. Though I can ensure that only the dialog query of the current addressed / controlling group is fetched through the use of  onQueryUpdated.

anehx: Let's just fire a query that fetches exactly what we need.. We don't really need to use "refetching".

luytena: Ok, I'll rewrite it :+1:

PR #2177, opened 2022-10-26
feat(sz): configurable navigation icons

luytena:


packages/distribution/addon/utils/inquiry-status.js
```diff
@@ -41,16 +43,49 @@ function decorator(
...
+                  ...config,
+                  color: config.color[inquiryType] ?? config.color,
+                };
+              })
+              .filter(Boolean)[0]
+          : null;
+
+      if (inquiryAnswerStatus) {
+        return inquiryAnswerStatus;
+      }
```
anehx, line 82: Where's the config part for this?

anehx (changes requested):
The tests for the inquiry status util absolutely needs to be extended. Otherwise nice work!

packages/distribution/addon/utils/inquiry-status.js
luytena: Good point, I'll add one for the "confirm-inquiry-answer" work-item.

luytena: Added

luytena:
> The tests for the inquiry status util absolutely needs to be extended. Otherwise nice work!

I extended the necessary tests

anehx (approved):
Awesome work! :tada:

PR #2185, opened 2022-11-01
fix(distribution): fix various inquiry permissions

anehx:
- Only allow editing sent inquiries if sending would be allowed
- Make sure answer form can't be edited without a ready complete workflow form work item
- Refactor inquiry ability for better readability

### DIFF
--- a/packages/distribution/addon/abilities/inquiry.js
+++ b/packages/distribution/addon/abilities/inquiry.js
@@ -4,101 +4,131 @@ import { DateTime } from "luxon";
 
 import config from "@projectcaluma/ember-distribution/config";
 
+/**
+ * This class contains all permission definitions for inquiries. To improve
+ * performance there are a few helpers and rules for optimal permission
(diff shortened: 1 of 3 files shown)


packages/distribution/addon/abilities/inquiry.js
```diff
@@ -4,101 +4,129 @@ import { DateTime } from "luxon";
...
+      .map(String)
+      .includes(String(this.calumaOptions.currentGroupId));
+  }
+
+  get canEdit() {
+    return (
+      this.hasBasePermission &&
+      // Since editing in the status ready has the same character as sending an
+      // inquiry, we need to make sure that permission would be given
+      (this.isSuspended || (this.isReady && this.canSend)) &&
```
luytena, line 62: this.canSend will only be true if the inquiry is suspended, no?

anehx: fixed

luytena (approved):

PR #2186, opened 2022-11-01
fix(distribution): fetch inquiry child-case document modifiedContentAt in dialog query

luytena:


packages/distribution/addon/gql/queries/inquiry-dialog.graphql
```diff
@@ -24,6 +24,7 @@ fragment DialogInquiry on WorkItem {
     ...InquiryAnswerButtons
     document {
       id
+      modifiedContentAt
```
anehx, line 27: Let's put this into the `InquiryAnswer` fragment

anehx (approved):

PR #2192, opened 2022-11-07
fix(distribution): custom edit inquiry permission

luytena:
Include custom edit inquiry permission in inquiry
abilities and improve inquiry views for users
without edit permission.

### DIFF
--- a/packages/distribution/addon/abilities/inquiry.js
+++ b/packages/distribution/addon/abilities/inquiry.js
@@ -57,12 +57,9 @@ export default class InquiryAbility extends Ability {
   get canEdit() {
     return (
       this.hasBasePermission &&
-      // Since editing in the status ready has the same character as sending an
-      // inquiry, we need to make sure that permission would be given
-      (this.isSuspended ||
-        (this.isReady &&
-          this.hasCustomPermission("sendInquiry", this.model))) &&
-      this.isControlling
+      (this.isSuspended || this.isReady) &&
+      this.isControlling &&
+      this.hasCustomPermission("editInquiry", this.model)
     );
   }
 
--- a/packages/distribution/addon/components/cd-inquiry-dialog/inquiry-part.js
+++ b/packages/distribution/addon/components/cd-inquiry-dialog/inquiry-part.js
@@ -33,10 +33,7 @@ export default class CdInquiryDialogInquiryPartComponent extends Component {
 
     const inquiry = this.args.inquiry;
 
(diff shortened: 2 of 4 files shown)


packages/distribution/addon/abilities/inquiry.js
```diff
@@ -59,7 +59,8 @@ export default class InquiryAbility extends Ability {
       this.hasBasePermission &&
       // Since editing in the status ready has the same character as sending an
       // inquiry, we need to make sure that permission would be given
-      (this.isSuspended ||
+      ((this.isSuspended &&
+        this.hasCustomPermission("editInquiry", this.model)) ||
         (this.isReady &&
           this.hasCustomPermission("sendInquiry", this.model))) &&
```
anehx, line 65: If we have a custom permission for it now we should drop the send inquiry custom permission logic in here.

packages/distribution/addon/components/cd-inquiry-dialog/inquiry-part.js
```diff
@@ -33,10 +33,7 @@ export default class CdInquiryDialogInquiryPartComponent extends Component {
 
     const inquiry = this.args.inquiry;
 
-    if (
-      inquiry.status === "SUSPENDED" &&
-      this.abilities.can("edit inquiry", inquiry)
```
anehx, line 38: Why did you drop this?

anehx (changes requested):

packages/distribution/addon/components/cd-inquiry-dialog/inquiry-part.js
luytena: Because without edit permission it is difficult to distinguish whether an inquiry was already sent or not.

anehx: Ah makes sense. I added this because I didn't want the creator of the inquiry to see this. But they only see the answer once it's completed - so NVM!

packages/distribution/addon/abilities/inquiry.js
luytena: Couldn't the permission still be variable depending on the status?

anehx: If it has to be it can be handled in the custom permission

luytena: Fair point, I updated it.

PR #2205, opened 2022-11-11
feat(form): use cf-content context for fetching and saving dynamic questions

anehx:
Needs https://github.com/projectcaluma/caluma/pull/1917


anehx:
Offline approved by @czosel

PR #2221, opened 2022-11-24
fix(cfb): fix min, max being invalid when empty

Yelinz:
## Description

Integer questions currently can't be saved without setting a min and max value, because the validation deems undefined an invalid value

### DIFF
--- a/packages/form-builder/addon/validations/question.js
+++ b/packages/form-builder/addon/validations/question.js
@@ -27,14 +27,14 @@ export default {
     validateType("IntegerQuestion", false),
     and(
       validateNumber({ allowBlank: true, integer: true }),
-      validateGtLt({ lt: "integerMaxValue" })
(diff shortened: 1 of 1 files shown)


anehx (changes requested):
Why didn't you do that in the validation definitions for the models? We might not want this everywhere..

Yelinz:
Changed

packages/form-builder/addon/validators/gt-lt.js
```diff
@@ -15,12 +15,6 @@ export default function validateGtLt(options = {}) {
...
     }, {});
 
-    return validateNumber(parsedOptions)(
-      key,
-      newValue,
-      oldValue,
-      changes,
-      content
-    );
+    return validateNumber(parsedOptions)(key, newValue);
```
anehx, line 18: Why this?

anehx (approved):

PR #2311, opened 2023-02-08
fix(form): use alt input with flatpickr

### LINKED ISSUE #2233: Datepicker issues
When trying to manually write a date, the input field bugs out and doesn't set the written date but a completely different one.

This behavior is caused by us using `formatDate` without changing the default `dateFormat`.
Because the default format is `Y-m-d` you can enter dates in that format, but trying to write dates in the displayed way doesn't work.

Either we provide a probably configurable `dateFormat` which then allows typed input or completely disable typed input

Yelinz:
## Description

Fix various bugs with the date input

Fixes #2233

### DIFF
--- a/packages/-ember-caluma/config/environment.js
+++ b/packages/-ember-caluma/config/environment.js
@@ -25,6 +25,12 @@ module.exports = function (environment) {
 
     "ember-caluma": {
       powerSelectEnableSearchLimit: 2,
+      FLATPICKR_DATE_FORMAT: {
+        de: "d.m.Y",
+        fr: "d.m.Y",
+        en: "m/d/Y",
+      },
+      FLATPICKR_DATE_FORMAT_DEFAULT: "m/d/Y",
     },
   };
 
--- a/packages/form/addon/components/cf-field/input/date.js
+++ b/packages/form/addon/components/cf-field/input/date.js
@@ -1,3 +1,4 @@
+import { getOwner } from "@ember/application";
 import { action } from "@ember/object";
 import { inject as service } from "@ember/service";
 import Component from "@glimmer/component";
@@ -13,22 +14,17 @@ export default class CfFieldInputDateComponent extends Component {
     return this.intl.primaryLocale.split("-")[0];
   }
 
-  /*
-   * Extract the date format string for the current locale.
-   * For example en-us will return "m/d/Y" and de-ch will return "d.m.Y".
-   */
-  get dateFormat() {
-    const sample = this.intl.formatDate(
-      DateTime.fromISO("1970-12-31").toJSDate(),
-      {
-        day: "2-digit",
-        month: "2-digit",
-        year: "numeric",
-      }
-    );
+  get config() {
+    return getOwner(this).resolveRegistration("config:environment");
+  }
 
-    // refer to flatpickr documentation for formatting options
-    return sample.replace(31, "d").replace(12, "m").replace(1970, "Y");
+  get dateFormat() {
+    const {
+      FLATPICKR_DATE_FORMAT = {},
+      FLATPICKR_DATE_FORMAT_DEFAULT = "m/d/Y",
+    } = this.config["ember-caluma"] || {};
+
+    return FLATPICKR_DATE_FORMAT[this.locale] ?? FLATPICKR_DATE_FORMAT_DEFAULT;
   }
 
   @action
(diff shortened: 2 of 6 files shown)


packages/form/addon/components/cf-field/input/date.js
```diff
@@ -1,27 +1,56 @@
...
+      {
+        day: "2-digit",
+        month: "2-digit",
+        year: "numeric",
+      }
+    );
+
+    // refer to flatpickr documentation for formatting options
+    return sample.replace(31, "d").replace(12, "m").replace(1970, "Y");
+  }
```
luytena, line 32: Is there any other way to get the format string based on the current locale?

Yelinz: I haven't found a better way, do you have an alternative?

luytena: I did some research and I didn't find another way to do this generically either. Since we only really support three locales (de, fr, en) I would suggest simply defining the format in the `environment.js` and accessing the format through a constant mapping such as:

 ```suggestion
  get dateFormat() {
        return ENV.APP.LOCALE_DATE_FORMAT[this.locale];
  }
```
Should another locale be supported in the future we can add another format. What do you think?

Comment (Yelinz) [reply packages/form/addon/components/cf-field/input/date.js]: Put it into the env

File: packages/form/addon/components/cf-field/input/date.js [hunk 2]
```diff
@@ -1,8 +1,10 @@
+import { getOwner } from "@ember/application";
 import { action } from "@ember/object";
 import { inject as service } from "@ember/service";
 import Component from "@glimmer/component";
 import { tracked } from "@glimmer/tracking";
 import { DateTime } from "luxon";
+// import ENV from "@projectcaluma/ember-form/config/environment";
```
luytena, line 7: Can be removed

packages/form/addon/components/cf-field/input/date.js
```diff
@@ -13,22 +15,17 @@ export default class CfFieldInputDateComponent extends Component {
...
-      {
-        day: "2-digit",
-        month: "2-digit",
-        year: "numeric",
-      }
-    );
-
-    // refer to flatpickr documentation for formatting options
-    return sample.replace(31, "d").replace(12, "m").replace(1970, "Y");
+    const config = getOwner(this).resolveRegistration("config:environment");
```
luytena, line 19: I would move the config resolving in a separate getter, since it doesn't change along with the locale.
```
get config() {
    return getOwner(this).resolveRegistration("config:environment");
  }
```

packages/-ember-caluma/config/environment.js
```diff
@@ -25,6 +25,12 @@ module.exports = function (environment) {
 
     "ember-caluma": {
       powerSelectEnableSearchLimit: 2,
+      FLATPICKR_DATE_FORMAT: {
+        de: "d.m.Y",
+        fr: "d.m.Y",
+        en: "m/d/Y",
+      },
+      FLATPICKR_DATE_FORMAT_DEFAULT: "m/d/Y",
```
luytena, line 33: Is the default locale English?

packages/form/addon/components/cf-field/input/date.js
```diff
@@ -1,27 +1,53 @@
...
+    const {
+      FLATPICKR_DATE_FORMAT = {},
+      FLATPICKR_DATE_FORMAT_DEFAULT = "m/d/Y",
+    } = config["ember-caluma"] || {};
+
+    if (!FLATPICKR_DATE_FORMAT[this.locale]) {
+      return FLATPICKR_DATE_FORMAT_DEFAULT;
+    }
+    return FLATPICKR_DATE_FORMAT[this.locale];
```
luytena, line 28:
```suggestion
    return FLATPICKR_DATE_FORMAT[this.locale] ?? FLATPICKR_DATE_FORMAT_DEFAULT;
```

packages/-ember-caluma/config/environment.js
Yelinz: Yes

packages/form/addon/components/cf-field/input/date.js
Yelinz: Done

packages/form/addon/components/cf-field/input/date.js
Yelinz: Done

packages/form/addon/components/cf-field/input/date.js
Yelinz: Done

luytena (approved):
LGTM :100:

PR #2317, opened 2023-02-16
fix(form): use defaults to mimic previous behaviour

Yelinz:
without settings defaults it would always choose the en format, when the options arent configured


luytena (commented):
Why isn't the default taken from the ember-caluma `environment.js`?

Yelinz:
The -ember-caluma environment is used for the documentation app

luytena:
What happens if we define it in the environment.js of ember-form?

Yelinz:
Its an addon therefore it doesnt have an environment

luytena (approved):

PR #2334, opened 2023-03-02
fix(analytics): fetch sorted analytics table for watchquery

Yelinz:
## Description

sorting resets when switching from preview to fields otherwise


velrest:
Does this have a performance impact somewhere?

Yelinz:
> Does this have a performance impact somewhere?

The query is only used in the reordering function

velrest (approved):

PR #2342, opened 2023-03-09
fix(form-builder): allow widget override to be removed

Yelinz:
## Description

currently once a override widget is selected, it cant be removed unless meta is edited


anehx (approved):
Nice catch!

PR #2343, opened 2023-03-09
feat(form-builder): add meta validation

Yelinz:
## Description
 add meta validation
code editor returns a string when its not a valid javascript object

### DIFF
--- a/packages/form-builder/addon/validators/meta.js
+++ b/packages/form-builder/addon/validators/meta.js
@@ -1,0 +1,13 @@
+export default function validateMeta() {
+  return (key, newValue) => {
+    // only allow objects {}
+    if (
+      typeof newValue === "object" &&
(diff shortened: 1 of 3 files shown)


packages/form-builder/addon/validators/meta.js
```diff
@@ -0,0 +1,13 @@
...
+  return (key, newValue) => {
+    // only allow objects {}
+    if (
+      typeof newValue === "object" &&
+      !Array.isArray(newValue) &&
+      newValue !== null
+    ) {
+      return true;
+    }
+    return false;
```
czosel, line 11:
```suggestion
    return (
      typeof newValue === "object" &&
      !Array.isArray(newValue) &&
      newValue !== null
    );
```

czosel (approved):

PR #2350, opened 2023-03-16
fix(analytics): fix table reordering breaking the table

Yelinz:
replace uikit sortable with ember-sortable

With uikit sortable the table can break when dragging to specific locations
Peek 2023-03-17 10-17

With ember-sortable it works and uses modifieres which is nice and relativly easily testable
Peek 2023-03-17 11-37

### DIFF
--- a/packages/analytics/addon/components/ca-field-selector-list.hbs
+++ b/packages/analytics/addon/components/ca-field-selector-list.hbs
@@ -1,8 +1,4 @@
-<table
-  class="uk-table uk-table-divider uk-table-hover"
-  uk-overflow-auto
-  {{did-insert this.setupUIkit}}
->
+<table class="uk-table uk-table-divider uk-table-hover uk-overflow-auto">
   <thead>
     <tr>
       <th></th>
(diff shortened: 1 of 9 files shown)


Yelinz:
Not sure why distribution tests are breaking, running them individually locally is passing

packages/analytics/addon/components/ca-field-selector-list.hbs
```diff
@@ -1,8 +1,4 @@
-<table
-  class="uk-table uk-table-divider uk-table-hover"
-  uk-overflow-auto
-  {{did-insert this.setupUIkit}}
->
+<table class="uk-table uk-table-divider uk-table-hover" uk-overflow-auto>
```
velrest, line 1: I think `uk-overflow-auto` is a class https://getuikit.com/docs/utility#overflow

velrest (commented):
Otherwise :+1:

packages/analytics/addon/components/ca-field-selector-list.hbs
Yelinz: Yes, changed it.
The tests are also breaking on main... something changed in the floating dependencies

velrest (approved):

PR #2362, opened 2023-03-27
fix: load the correct widget override options for a question

StephanH90:
## Description

Currently, the options for widget overrides are not being loaded correctly when changing the question type. This means that any widget override that has a question type specified is not available.

### correct behaviour (fixed):
[Screencast from 27.03.2023 16:42:30.webm](https://user-images.githubusercontent.com/88476449/227975707-d678c4fa-6cfb-40f8-a972-92d1276ddadd.webm)

### incorrect behaviour (before):
[Screencast from 27.03.2023 16:45:04.webm](https://user-images.githubusercontent.com/88476449/227975674-f9ff7531-6f2a-4410-970b-e14d4d35a222.webm)

## DISCLAIMER

I think the issue is that we are creating the changeset in a getter:

```javascript
get changeset () {
  return new Changeset(this.model, ...)
}
```

which messes up the reactivity when trying to watch changes on the changeset (I think?). Instead, I made the changeset a tracked property and initialise it after the data is loaded. However, this might mess up some other stuff, so I hope there is a cleaner solution.

### DIFF
--- a/packages/form-builder/addon/components/cfb-form-editor/question.hbs
+++ b/packages/form-builder/addon/components/cfb-form-editor/question.hbs
@@ -3,7 +3,7 @@
     <div class="uk-flex uk-flex-center uk-flex-middle uk-height-small">
       <UkSpinner @ratio={{2}} />
     </div>
-  {{else if this.model}}
+  {{else if this.changeset}}
     <ValidatedForm
       @model={{this.changeset}}
       @on-submit={{perform this.submit}}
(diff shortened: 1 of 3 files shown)


packages/form-builder/addon/components/cfb-form-editor/question.hbs
```diff
@@ -3,7 +3,7 @@
     <div class="uk-flex uk-flex-center uk-flex-middle uk-height-small">
       <UkSpinner @ratio={{2}} />
     </div>
-  {{else if this.model}}
+  {{else if (and this.model this.changeset)}}
```
anehx, line 6:
```suggestion
  {{else if this.changeset}}
```
Should be enough, no?

anehx (changes requested):

anehx (approved):

PR #2381, opened 2023-04-24
feat(form): add data attributes for better field identification

derrabauke:
## Description

We want to customize the layout of our caluma form. Therefore we have to identify the form field component, especially the wrapping `div`. Currently only the inputs are contextualized with the field/question id. Since CSS `:has` is not supported in Firefox yet, we have to mark the wrapping element as well.

image

### DIFF
--- a/packages/form/addon/components/cf-field.hbs
+++ b/packages/form/addon/components/cf-field.hbs
@@ -5,6 +5,8 @@
         (and @disabled (has-question-type @field.question 'action-button'))
         'uk-hidden'
       }}"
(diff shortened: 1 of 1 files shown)


derrabauke:
> Hm, the slug is already in the `name` attribute and the type in the `type` attribute.. What use case isn't covered by the existing code?

The wrapping div has no attributes besides the `uk-margin` class. How should we target them, if not with extra attributes?

anehx:
Fair enough.

anehx (approved):

PR #2387, opened 2023-04-27
chore(release): fix release in CI

anehx:
`release-it` doesn't work properly with `pnpm` so we dropped it in favor of our normal release tool `semantic-release`. However, that doesn't really work with monorepos and pnpm, so I added a little bit of customization:

1. The version bump is done via a node script since there is no good `semantic-release` plugin to do that in monorepos
2. The publish step is done via pnpm CLI to make sure the cross dependency versions are properly resolved

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,10 +2,6 @@ name: Release
 
 on:
   workflow_dispatch:
-    inputs:
-      prerelease:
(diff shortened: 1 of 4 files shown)


derrabauke (approved):
:sparkles: nice :sparkles:

derrabauke (approved):

PR #2489, opened 2023-08-10
feat: copy forms

StephanH90:
## Description

Allows you to copy forms from the form builder.

image

image

image

image

### DIFF
--- a/packages/form-builder/addon/components/cfb-form-list/copy-modal.hbs
+++ b/packages/form-builder/addon/components/cfb-form-list/copy-modal.hbs
@@ -1,0 +1,49 @@
+{{#if this.visible}}
+  <UkModal
+    @visible={{this.visible}}
+    @onHide={{fn (mut this.visible) false}}
+    data-test-copy-form-modal
+    as |modal|
+  >
+    <ValidatedForm
+      @model={{this.changeset}}
+      @on-submit={{perform this.submit}}
+      as |f|
+    >
+      <modal.header>
+        <h2 class="uk-modal-title">
+          {{t "caluma.form-builder.copy-modal.title" form=@item.name}}
+        </h2>
+      </modal.header>
+      <modal.body>
+        <f.input
+          @name="name"
+          @label={{t "caluma.form-builder.copy-modal.name.label"}}
+          @hint={{t "caluma.form-builder.copy-modal.name.hint" name=@item.name}}
+          data-test-copy-modal-input-name
+        />
+        <f.input
+          @name="slug"
+          @label={{t "caluma.form-builder.copy-modal.slug.label"}}
+          @hint={{t "caluma.form-builder.copy-modal.slug.hint" slug=@item.slug}}
+          data-test-copy-modal-input-slug
+        />
+      </modal.body>
+      <modal.footer class="uk-text-right">
+        <f.submit
+          @loading={{this.submit.isRunning}}
+          @disabled={{or
+            this.submit.isRunning
+            (not this.changeset.isValid)
+            (eq @item.slug this.changeset.slug)
+          }}
+          data-test-copy-form-submit
+        >
+          {{t "caluma.form-builder.copy-modal.submit"}}
+        </f.submit>
+      </modal.footer>
+    </ValidatedForm>
+  </UkModal>
+{{/if}}
+
+{{yield (hash toggle=this.toggle)}}
--- a/packages/form-builder/addon/components/cfb-form-list/copy-modal.js
+++ b/packages/form-builder/addon/components/cfb-form-list/copy-modal.js
@@ -1,0 +1,70 @@
+import { action } from "@ember/object";
+import { inject as service } from "@ember/service";
+import Component from "@glimmer/component";
+import { tracked } from "@glimmer/tracking";
+import { queryManager } from "ember-apollo-client";
+import { Changeset } from "ember-changeset";
+import lookupValidator from "ember-changeset-validations";
+import { dropTask } from "ember-concurrency";
+
+import copyFormMutation from "@projectcaluma/ember-form-builder/gql/mutations/copy-form.graphql";
+import validations from "@projectcaluma/ember-form-builder/validations/form";
+
+export default class componentsCfbFormItemListCopyModal extends Component {
+  @queryManager apollo;
+  @service notification;
+  @service router;
+  @service intl;
+  @tracked visible = false;
+
+  constructor(owner, args) {
+    super(owner, args);
+    this.changeset = Changeset(
+      this.args.item,
+      lookupValidator(validations),
+      validations,
+    );
+  }
+
+  @action
+  toggle() {
+    this.visible = !this.visible;
+  }
+
+  @dropTask
+  *submit(changeset) {
+    try {
+      const form = yield this.apollo.mutate(
+        {
+          mutation: copyFormMutation,
+          variables: {
+            input: {
+              source: this.args.item.slug,
+              name: changeset.name,
+              slug: changeset.slug,
+            },
+          },
+        },
+        "copyForm.form",
+      );
+
+      this.notification.success(
+        this.intl.t(
+          `caluma.form-builder.notification.form.${
+            this.args.slug ? "save" : "create"
+          }.success`,
+        ),
+      );
+
+      this.router.transitionTo("edit", form.slug);
+    } catch (e) {
+      this.notification.danger(
+        this.intl.t(
+          `caluma.form-builder.notification.form.${
+            this.args.slug ? "save" : "create"
+          }.error`,
+        ),
+      );
+    }
+  }
+}
--- a/packages/form-builder/addon/components/cfb-slug-input.hbs
+++ b/packages/form-builder/addon/components/cfb-slug-input.hbs
@@ -14,6 +14,7 @@
         </span>
       {{/if}}
       <input
+        ...attributes
         id={{@inputId}}
         name={{or @inputName @name}}
         type="text"
--- a/packages/form-builder/addon/components/cfb-slug-input.js
+++ b/packages/form-builder/addon/components/cfb-slug-input.js
@@ -33,7 +33,7 @@ export default class CfbSlugInputComponent extends Component {
   calculatePadding(element) {
     // Ensures that the element is already visible in combined use
     // with modal dialogs.
-    scheduleOnce("afterRender", this, this._calculatePadding, element);
+    scheduleOnce("afterRender", this, "_calculatePadding", element);
   }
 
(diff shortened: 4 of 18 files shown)


packages/form-builder/addon/gql/mutations/copy-form.graphql
```diff
@@ -0,0 +1,11 @@
+#import FormInfo from '../fragments/form-info.graphql'
+
+mutation CopyForm($input: CopyFormInput!) {
+  copyForm(input: $input) {
+    form {
+      id
+      ...FormInfo
```
anehx, line 7: Does this prevent the extra request after the redirect?

packages/form-builder/addon/components/cfb-form-list/copy-modal.hbs
```diff
@@ -0,0 +1,49 @@
+{{#if this.visible}}
+  <UkModal
+    @visible={{this.visible}}
```
anehx, line 3: This will probably give you some bad UX. I understand the reason for it, but if we just moved the button to the detail view, we wouldn't need to think about it..

packages/form-builder/addon/components/cfb-form-list/copy-modal.hbs
```diff
@@ -0,0 +1,49 @@
...
+          @label={{t "caluma.form-builder.copy-modal.name.label"}}
+          @hint={{t "caluma.form-builder.copy-modal.name.hint" name=@item.name}}
+          data-test-copy-modal-input-name
+        />
+        <f.input
+          @name="slug"
+          @label={{t "caluma.form-builder.copy-modal.slug.label"}}
+          @hint={{t "caluma.form-builder.copy-modal.slug.hint" slug=@item.slug}}
+          data-test-copy-modal-input-slug
+        />
```
anehx, line 30: You need to use the `cfb-slug-input` to make sure namespaces are handled

packages/form-builder/addon/components/cfb-form-list/copy-modal.hbs
```diff
@@ -0,0 +1,49 @@
...
+      </modal.body>
+      <modal.footer class="uk-text-right">
+        <f.submit
+          @loading={{this.submit.isRunning}}
+          @disabled={{or
+            this.submit.isRunning
+            (not this.changeset.isValid)
+            (eq @item.slug this.changeset.slug)
```
anehx, line 38: Validation should already cover that, no?

packages/form-builder/addon/components/cfb-form-list/copy-modal.js
```diff
@@ -0,0 +1,70 @@
...
+            },
+          },
+        },
+        "copyForm.form",
+      );
+
+      this.notification.success(
+        this.intl.t(
+          `caluma.form-builder.notification.form.${
+            this.args.slug ? "save" : "create"
```
anehx, line 54: There is no such argument

packages/form-builder/addon/components/cfb-form-list/copy-modal.js
```diff
@@ -0,0 +1,70 @@
...
+          }.success`,
+        ),
+      );
+
+      this.router.transitionTo("edit", form.slug);
+    } catch (e) {
+      this.notification.danger(
+        this.intl.t(
+          `caluma.form-builder.notification.form.${
+            this.args.slug ? "save" : "create"
```
anehx, line 64: Same here

packages/form-builder/translations/en.yaml
```diff
@@ -169,6 +169,16 @@ caluma:
...
+    copy-modal:
+      title: 'Copy "{form}"'
+      name:
+        label: "Name"
+        hint: "Original: {name}"
+      slug:
+        label: "Slug"
+        hint: "Original: {slug}"
+      submit: "Copy form"
+
```
anehx, line 181: Add french deepl translations please

anehx (changes requested):

derrabauke:
> I'd like to have this button on the form detail view as well [...]

Would you like it like this? 
image

Mockups welcome :sunglasses:

anehx:
> > I'd like to have this button on the form detail view as well [...]
> 
> Would you like it like this? image
> 
> Mockups welcome 😎

Yeah this seems nice. I'd name it just "Copy" but other than that perfect.

Also, IMO this is the only place this is needed. I don't see the use case in the list view..

packages/form-builder/addon/components/cfb-slug-input.hbs
```diff
@@ -14,6 +14,7 @@
         </span>
       {{/if}}
       <input
+        ...attributes
```
derrabauke, line 17: For splattering `data-test` attributes.

packages/form-builder/addon/components/cfb-slug-input.js
```diff
@@ -30,6 +31,16 @@ export default class CfbSlugInputComponent extends Component {
 
   @action
   calculatePadding(element) {
+    // Ensures that the element is already visible in combined use
+    // with modal dialogs.
+    scheduleOnce(
+      "afterRender",
+      this,
+      this._calculatePadding.bind(this, element),
```
derrabauke, line 39: To prevent `ember/no-incorrect-calls-with-inline-anonymous-functions` linting error.

anehx: Just pass "_calculatePadding" - this will take care of bindings.

anehx: To be precise:

```js
scheduleOnce("afterRender", this, "_calculatePadding", element)
```

packages/form-builder/addon/components/cfb-form-list/copy-modal.hbs
anehx: But the slug validation needs to check if there is already a form with this slug. So if this doesn't work then the validation is wrong.

packages/form-builder/addon/components/cfb-slug-input.js
```diff
@@ -30,6 +31,12 @@ export default class CfbSlugInputComponent extends Component {
 
   @action
   calculatePadding(element) {
+    // Ensures that the element is already visible in combined use
+    // with modal dialogs.
+    scheduleOnce("afterRender", this, this._calculatePadding, element);
```
anehx, line 36: Must be a string, otherwise context binding will be weird.
```suggestion
    scheduleOnce("afterRender", this, "_calculatePadding", element);
```

packages/form-builder/addon/gql/mutations/copy-form.graphql
anehx: Not addressed yet

packages/form-builder/addon/components/cfb-form-list/copy-modal.hbs
anehx: Not addressed yet

derrabauke: I've already moved the `{{#if this.visible}}` into the modal body around the `ValidatedForm`. Fixes the *bad UX* as the modal is animating properly and gives enough feedback for the tests. So I guess it's *OK* now?!

packages/form-builder/addon/gql/mutations/copy-form.graphql
derrabauke: Without the `FormInfo` fragment, we would not receive the `slug` which is necessary as a query param for the redirect. So yes, it's essential here.

packages/form-builder/addon/components/cfb-form-list/copy-modal.hbs
derrabauke: You were right! The validation was tricked, cause the form model already got an `id` property (that of the original form), so it skipped the slug validation. `fixed`

anehx: Nice :tada:

packages/form-builder/addon/gql/mutations/copy-form.graphql
anehx: But does it prevent loading the whole form after the redirect? If not, I'd suggest only fetching the slug (which we'd actually know without the response, just saying) as we only need that and not the whole form data.

**However,** if the prevents the query after we redirect to the new form then I'm all for it as we save ourselves one request :)

packages/form-builder/addon/components/cfb-slug-input.js
```diff
@@ -30,6 +31,12 @@ export default class CfbSlugInputComponent extends Component {
 
   @action
   calculatePadding(element) {
+    // Ensures that the element is already visible in combined use
+    // with modal dialogs.
+    scheduleOnce("afterRender", this, "_calculatePadding", element);
+  }
+
+  _calculatePadding(element) {
```
anehx, line 39: BTW just some random JS facts: if you really want this to be private, use `#calculatePadding`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

derrabauke: > BTW just some random JS facts: if you really want this to be private, use `#calculatePadding`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

Thanks. :santa: 

`scheduleOnce` can not resolve private methods provided as string reference. But we could provide it directly as a function. But then again your previous critic comes into effect. ("*Must be a string, otherwise context binding will be weird*"). I propose we stick with the pseudo private underscore oldsql naming.

packages/form-builder/addon/gql/mutations/copy-form.graphql
derrabauke: Makes no difference. I will delete the fragment from the mutation response. Save us 200 *bytes* :tada:

derrabauke:
@anehx any clue why [this test error](https://github.com/projectcaluma/ember-caluma/actions/runs/7277728691/job/19830421002?pr=2489#step:4:58) pops up now? Can not reproduce with `ember test -s` but with `ember test`. o.O

derrabauke:
Seems like calling `setupIntl(hooks, "en") with a locale is fixing it.

derrabauke:
@anehx this is getting stale. If you find a minute, could you have a look at it again?

packages/form-builder/addon/components/cfb-slug-input.js
anehx: TIL

anehx (approved):
Nice work @derrabauke & @StephanH90

PR #2515, opened 2023-09-20
fix(form): fix fast selection in checkboxes

anehx:
This fixes an issue where the value of a checkbox question was not properly updated to the answer model and therefore the backend. This happened because the save task uses a timeout to debounce value saving. As the update function of the component always used the value from the user model this caused a race condition if the user clicked the options too fast.


Yelinz (approved):

czosel (approved):
Nice, even with a test! :+1:

PR #2525, opened 2023-10-04
fix: add ember-data version check to check if calling `.toArray` on ember-data arrays is deprecated

StephanH90:
Calling `.toArray` on ember-data arrays is deprecated since `ember-data@4.7.0`. Therefore we make the `.toArray` call conditional.

### DIFF
--- a/packages/core/addon/services/-scheduler.js
+++ b/packages/core/addon/services/-scheduler.js
@@ -2,9 +2,12 @@ import { assert } from "@ember/debug";
 import { once } from "@ember/runloop";
 import Service, { inject as service } from "@ember/service";
 import { camelize } from "@ember/string";
+import { dependencySatisfies } from "@embroider/macros";
 import { tracked } from "@glimmer/tracking";
 import { task } from "ember-concurrency";
 import { pluralize } from "ember-inflector";
+
+const toArrayIsDeprecated = dependencySatisfies("ember-data", "^4.7.0");
 
 /**
  * Decorator to define a type resolver in the scheduler service.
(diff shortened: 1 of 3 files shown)


packages/core/addon/services/-scheduler.js
```diff
@@ -37,7 +40,9 @@ function typeResolver(type) {
       ? yield this.calumaOptions[methodName]?.(uncachedIdentifiers)
       : [];
 
-    const allResults = [...cached, ...(result?.toArray?.() ?? result ?? [])];
+    const allResults = toArrayIsDeprecated
+      ? [...cached, ...(result ?? [])]
+      : [...cached, ...(result?.toArray?.() ?? result ?? [])];
```
czosel, line 45: In which situation would we still need the old syntax?

StephanH90: When using `ember-caluma` with `ember-data < 4.7`.

Otherwise, we have to do a major version and specify `ember-data > 4.7` as a dependency.

czosel: Would be interesting to know since when exactly the new syntax is supported - but I wasn't able to find it in a quick search. Thanks!

packages/core/package.json
```diff
@@ -15,6 +15,7 @@
   "dependencies": {
     "@apollo/client": "^3.7.17",
     "@ember/string": "^3.1.1",
+    "@embroider/macros": "1.13.0",
```
anehx, line 18:
```suggestion
    "@embroider/macros": "^1.13.0",
```

This is already used in the distribution package and should be the same

anehx (changes requested):

packages/core/package.json
StephanH90: done

anehx (changes requested):
The changes in the lockfile seem very weird and don't make any sense to me. Could you reset the lockfile to `main` and run `pnpm i` again?

StephanH90:
I had to downgrade my pnpm to 8.6.0 (I was on 8.8.0 before). Now the lockfile looks right.

anehx (approved):

PR #2530, opened 2023-10-06
feat: enable feature to show all services when selecting services for…

StephanH90:
Enables a new way to select services to be invited to an inquiry:
image
image
image

- Individual sections can be shown/hidden
- When searching all the sections are shown automatically:


czosel (approved):
LGTM!

PR #2549, opened 2023-11-20
feat(check-question): add reset button for optional  question

amchangama:
## Description

Adds a reset button after non-required inputs of a `choice` question type.

## Preview

image


derrabauke (approved):
I've added a little PR description+preview. Code looks good :+1:

packages/form/addon/components/cf-field/input/radio.hbs
```diff
@@ -22,4 +23,9 @@
       {{option.label}}
     {{/if}}
   </label>
-{{/each}}
\ No newline at end of file
+{{/each}}
+{{#if (and (eq @field.question.raw.isRequired "false") this.checkedValue)}}
+  <button {{on "click" this.resetSelected}} class="reset" type="button">{{t
+      "caluma.form.power-select.reset"
+    }}</button>
```
anehx, line 30: Let's just use a link so we can drop the custom styling.

packages/form/addon/components/cf-field/input/radio.js
```diff
@@ -0,0 +1,26 @@
+import { action } from "@ember/object";
+import Component from "@glimmer/component";
+import { tracked } from "@glimmer/tracking";
+
+export default class CfFieldInputRadioComponent extends Component {
+  @tracked checkedValue = this.args.field.answer.value;
+
+  @action
+  updatSelection({ target: { value, checked } }) {
```
czosel, line 9:
```suggestion
  updateSelection({ target: { value, checked } }) {
```

packages/form/addon/components/cf-field/input/radio.js
```diff
@@ -0,0 +1,26 @@
...
+    if (this.args.field.question.raw.isRequired === "false" && checked) {
+      this.checkedValue = value;
+    } else if (this.args.field.question.raw.isRequired === "true") {
+      this.args.onSave(value);
+    } else {
+      this.args.onSave(null);
+    }
+  }
```
czosel, line 17: Why is this method needed? Seems to me that we only need `resetSelected`.

packages/form/tests/integration/components/cf-field/input/radio-test.js
```diff
@@ -87,18 +87,31 @@ module("Integration | Component | cf-field/input/radio", function (hooks) {
...
-    options=(array (hash slug="option-1" label="Option 1"))
-    question=(hash __typename="ChoiceQuestion")
-  }}
-/>`);
+    await render(
+      hbs`<CfField::Input::Radio @onSave={{this.save}} @field={{this.field}} />`,
+    );
 
     await click("label:nth-of-type(1) input");
+    assert.equal(true, true);
```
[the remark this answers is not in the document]
derrabauke: `resolved`

packages/form/addon/components/cf-field/input/radio.js
derrabauke: `reverted`

packages/form/addon/components/cf-field/input/radio.hbs
```diff
@@ -22,4 +22,11 @@
       {{option.label}}
     {{/if}}
   </label>
-{{/each}}
\ No newline at end of file
+{{/each}}
+{{#if (and (eq @field.question.raw.isRequired "false") @field.answer.value)}}
```
anehx, line 26:
```suggestion
{{#if (and @field.optional @field.answer.value)}}
```

packages/form/addon/components/cf-field/input/radio.hbs
```diff
@@ -22,4 +22,11 @@
...
+  <br />
+  <a data-test-radio-reset href="#" {{on "click" (fn @onSave null)}}>{{t
+      "caluma.form.power-select.reset"
+    }}</a>
```
anehx, line 30: Weird formatting

packages/form/addon/components/cf-field/input/radio.hbs
```diff
@@ -22,4 +22,11 @@
       {{option.label}}
     {{/if}}
   </label>
-{{/each}}
\ No newline at end of file
+{{/each}}
+{{#if (and (eq @field.question.raw.isRequired "false") @field.answer.value)}}
+  <br />
+  <a data-test-radio-reset href="#" {{on "click" (fn @onSave null)}}>{{t
```
anehx, line 28: This will cause a `#` being appended to the URL which we don't want. Please add an intermediate action which includes `event.preventDefault()`

packages/form/package.json
```diff
@@ -35,6 +35,7 @@
     "ember-math-helpers": "^4.0.0",
     "ember-power-select": "^7.2.0",
     "ember-resources": "^6.4.2",
+    "ember-truth-helpers": "^4.0.3",
```
anehx, line 38: As we only need this in one place I'd suggest putting the logic into the JS file to avoid unnecessary dependencies

anehx (changes requested):

packages/form/addon/components/cf-field/input/radio.hbs
derrabauke: `applied`

packages/form/addon/components/cf-field/input/radio.hbs
derrabauke: Can't change what prettier is doin' here :man_shrugging:

packages/form/addon/components/cf-field/input/radio.hbs
derrabauke: `done`

packages/form/package.json
derrabauke: I totally see your point, but I wanted to spare the `component.js` file (it's now there anyways) and saw that the `workflow` package has it as well. Further I think the [`action-button.hbs`](https://github.com/projectcaluma/ember-caluma/blob/649c5d8524bf930a78c9395e1516b38b63c62316/packages/form/addon/components/cf-field/input/action-button.hbs#L33) is using it as well, or am I missing something there?

packages/form/addon/components/cf-field/input/radio.hbs
anehx: Yeah you can:
```hbs
<a data-test-radio-reset href="#" {{on "click" (fn @onSave null)}}>
  {{t "caluma.form.power-select.reset"}}
</a>
```
Prettier is not always prettier.

anehx:
@derrabauke You forgot to push.. There are no changes after my review.

packages/form/package.json
anehx: True. There are some other places where it's used as well. I withdraw my comment :wink:

anehx: **But:** remember to add it in the blueprint as the host app must install it because it's used in templates

derrabauke:
> @derrabauke You forgot to push.. There are no changes after my review.

Not forgotten, but I was tangled in confusion how the `ember-truth-helpers` references get resolved in the app and if it should be a dependency or not. :exploding_head:

packages/form/package.json
derrabauke: `done`

packages/form/addon/components/cf-field/input/radio.hbs
```diff
@@ -22,4 +22,10 @@
       {{option.label}}
     {{/if}}
   </label>
-{{/each}}
\ No newline at end of file
+{{/each}}
+{{#if (and @field.optional @field.answer.value)}}
+  <br />
```
anehx, line 27: Can we replace this with `uk-margin-small-top` on the link?

anehx:
Alright, one final nitpick then it is ready to go :tada: Sorry @derrabauke :wink:

packages/form/addon/components/cf-field/input/radio.hbs
derrabauke: Sure. I'll give it a shot and push. brb

derrabauke: `a` is an inline element. So we would have to add `uk-display-block` at least. Extra margin to that does not look good (to much of a gap between options and reset) *imho*. 

That's how it looks with `uk-display-block` only:
image

And that's with the extra `uk-margin-small-top`:
image

anehx: Let's keep it like it is then. Thanks for investigating.

anehx (approved):

packages/form/addon/components/cf-field/input/radio.hbs
derrabauke: Well, it doesn't look to bad paired with follow up questions. And it has the same margin then like *hints*. 
image

Will add the small margin. ;)

derrabauke: Ok to late :D

PR #2567, opened 2023-12-08
fix(distribution): improve text when no action is available

Yelinz:
## Description

The current text on the distribution index page is misleading when no actions are available

image

### DIFF
--- a/packages/distribution/translations/de.yaml
+++ b/packages/distribution/translations/de.yaml
@@ -1,6 +1,7 @@
 caluma:
   distribution:
     empty: "Es wurden noch keine Anfragen erstellt."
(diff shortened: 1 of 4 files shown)


packages/distribution/translations/de.yaml
```diff
@@ -1,6 +1,7 @@
 caluma:
   distribution:
     empty: "Es wurden noch keine Anfragen erstellt."
+    no-actions: "Keine Aktionen im aktuellen Zustand möglich."
```
czosel, line 4:
```suggestion
    no-actions: "Im aktuellen Zustand stehen keine Aktionen zur Verfügung."
```

czosel (approved):
Just a nitpick, LGTM

PR #2584, opened 2023-12-19
fix(form-builder): fix cursor issues in code editor fields

anehx:


anehx:
Offline-approved by @czosel

PR #2610, opened 2024-01-23
feat(form): add widget override for number separator

anehx:
image

Peek 2024-01-23 12-26

### DIFF
--- a/packages/form/addon/components/cf-field/input/number-separator.hbs
+++ b/packages/form/addon/components/cf-field/input/number-separator.hbs
@@ -1,0 +1,12 @@
+<input
+  type="text"
+  class="uk-input
+    {{if @field.isInvalid 'uk-form-danger'}}
+    {{if this.disabled 'uk-disabled'}}"
+  name={{@field.pk}}
+  id={{@field.pk}}
+  value={{this.displayValue}}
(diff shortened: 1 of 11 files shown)


packages/form/addon/components/cf-field/input/number-separator.hbs
```diff
@@ -0,0 +1,13 @@
...
+  type="text"
+  class="uk-input
+    {{if @field.isInvalid 'uk-form-danger'}}
+    {{if this.disabled 'uk-disabled'}}"
+  name={{@field.pk}}
+  id={{@field.pk}}
+  value={{this.displayValue}}
+  placeholder={{@field.question.raw.placeholder}}
+  readonly={{this.disabled}}
+  pattern={{this.pattern}}
```
derrabauke, line 11: There is no `this.pattern` in the component. Should it be an argument?

derrabauke (changes requested):
One minor nitpick

packages/form/addon/components/cf-field/input/number-separator.hbs
anehx: Leftover from something I tried. Removed it.

derrabauke (approved):
:+1:

PR #2611, opened 2024-01-25
fix(form-builder): don't refetch the forms for "used by" modal on changeset update

StephanH90:
## Description

Currently a change in the question triggers a re-fetching of all the forms where this question is used. (Maybe because changeset replaces the tracked object instead of updating its property? Not sure)

To prevent this we simply assign the result of the graphql query to a tracked array and re-use this when the trackedFunction is called again.


derrabauke (approved):
LGTM :+1:

StephanH90:
@anehx btw I can't merge this because of the flaky test.

PR #2622, opened 2024-01-30
fix: add optional peer dep on ember data

czosel:
Without this, the `dependencyStatisfies` embroider macro won't work -
see https://github.com/embroider-build/embroider/issues/1169#issuecomment-1081943925
for details.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -65,11 +65,17 @@
     "webpack": "5.90.0"
   },
   "peerDependencies": {
-    "ember-source": "^4.0.0"
+    "ember-source": "^4.0.0",
+    "ember-data": "*"
   },
   "dependenciesMeta": {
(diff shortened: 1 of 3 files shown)


StephanH90 (approved):

anehx (approved):

packages/core/package.json
```diff
@@ -65,13 +65,19 @@
     "webpack": "5.90.0"
   },
   "peerDependencies": {
-    "ember-source": "^4.0.0"
+    "ember-source": "^4.0.0",
+    "ember-data": ">=4.0.0"
```
anehx, line 69: Potentially even `*` as ember-caluma doesn't use it and doesn't care about the version of ember-data at all.

czosel: sure, fixed

Issue #2514, opened 2023-09-15
Flakey test in ember-form


Yelinz:
Sometimes fails, retry of test passes
```
not ok 29 Chrome 116.0 - [101 ms] - Integration | Component | cf-field/input/action-button: validates as configured
    ---
        actual: >
            
        expected: >
            validate
        stack: >
                at Assert.verifySteps (webpack://__ember_auto_import__/../../node_modules/.pnpm/qunit@2.19.4/node_modules/qunit/qunit/qunit.js?:283:51)
        negative: >
            false
        browser log: |
    ...
```
https://github.com/projectcaluma/ember-caluma/actions/runs/6170568375/job/16747149151
https://github.com/projectcaluma/ember-caluma/actions/runs/6183695072/job/16785971723

derrabauke:
:heavy_plus_sign: [that](https://github.com/projectcaluma/ember-caluma/actions/runs/7252527626/job/19757227879?pr=2467#step:4:85)

Yelinz:
Only fails on chrome, have never gotten it to fail on firefox.
Maybe we should just remove this test, as the functionality is already being tested multiple times in other tests. https://github.com/projectcaluma/ember-caluma/blob/ebcd822c65b8639f6449a702e74a8129f14f718d/packages/form/tests/integration/components/document-validity-test.js#L34-L54

Any opinion @anehx @derrabauke

anehx:
@Yelinz yeah, let's remove it so our pipelines have meaning again..

PR #2635, opened 2024-02-16
fix(cf-field-radio): hide reset button if form is disabled

MitanOmar:


derrabauke (approved):
:+1:

derrabauke:
The failing tests are the flaky ones mentioned here https://github.com/projectcaluma/ember-caluma/issues/2514

Feel free to merge

packages/form/addon/components/cf-field/input/radio.hbs
```diff
@@ -23,7 +23,7 @@
     {{/if}}
   </label>
 {{/each}}
-{{#if (and @field.optional @field.answer.value)}}
+{{#if (and (and @field.optional @field.answer.value) (not @disabled))}}
```
anehx, line 26:
```hbs
{{#if (and @field.optional @field.answer.value (not @disabled)}}
```

No need for the additional `and`

anehx (commented):
Looks good apart from a minor nitpick. @MitanOmar could you fix that?

packages/form/addon/components/cf-field/input/radio.hbs
MitanOmar: i was searching in the documentation if `and` accept more then two parameter, but i did not see it, all examples ware explaining with two options only, but it was in my mind, i will do it in new PR

anehx: Yeah I was very convinced that I'd previously used it with more than 2 parameters but had to check the code to be 100% sure: https://github.com/jmurphyau/ember-truth-helpers/blob/master/packages/ember-truth-helpers/src/helpers/and.ts

The helper loops over **all** parameters and returns `false` if any one of them is falsey.

anehx: Yeah I was very convinced that I'd previously used it with more than 2 parameters but had to check the code to be 100% sure: https://github.com/jmurphyau/ember-truth-helpers/blob/master/packages/ember-truth-helpers/src/helpers/and.ts

The helper loops over **all** parameters and returns `false` if any one of them is falsey.

PR #2645, opened 2024-03-07
fix(deps): update dependencies

anehx:
- Update general dependencies
- Update `ember-concurrency` to next major version v4
- Remove `ember-resources`:
  The tools from `ember-resources` that we used are being moved to
  `reactiveweb` and `ember-modify-based-class-resource`


Yelinz (approved):
didnt grep the codebase, but the needed migration steps were done :+1:

PR #2646, opened 2024-03-08
fix(perf): merge distribution navigation query into a single graph

anehx:


anehx:
@luytena @kaldras Do you think we should merge or discard this idea? IMO, if it's not necessary I'd keep it as is since the current solution is better to read and debug for devs.

PR #2663, opened 2024-03-18
fix(form): only pass locale string to flatpickr

anehx:


czosel:
Changes LGTM, just a question: is localeObject not supported anymore, or what was the motivation?

anehx:
> Changes LGTM, just a question: is localeObject not supported anymore, or what was the motivation?

It's still supported but using it was a workaround for a problem that is fixed now. This is much cleaner as the component will just import the needed locale object automatically.

PR #2689, opened 2024-04-15
fix(translations): change "optional" to "optionnelle"

StephanH90:
## Description

Changes "optional" in French to "optionnelle". That is, according to one of our French speaking clients, the correct translation.

### DIFF
--- a/packages/form/translations/fr.yaml
+++ b/packages/form/translations/fr.yaml
@@ -1,6 +1,6 @@
 caluma:
   form:
-    optional: "optional"
+    optional: "optionnelle"
     save: "sauvegarder"
     delete: "supprimer"
     edit: "modifier"


amchangama:
**optionnelle** is the feminine, using **optionnel** (masculin) is common if we are not sure of what were refering to(either feminine or masculine)

anehx:
> **optionnelle** is the feminine, using **optionnel** (masculin) is common if we are not sure of what were refering to(either feminine or masculine)

In this case "optional" refers to the question which is a feminine word: https://dictionnaire.lerobert.com/definition/question

So IMO this change is correct, right @amchangama ?

amchangama:
@anehx  yes it is correct if it refers to a question

anehx (approved):

PR #2691, opened 2024-04-19
feat(analytics): excel export of reports table

open-dynaMIX:
## Description

Export your analytics preview table as Excel. 

2024-04-19_11-47-54_1323x588

### DIFF
--- a/packages/analytics/addon/components/ca-report-preview.hbs
+++ b/packages/analytics/addon/components/ca-report-preview.hbs
@@ -4,7 +4,16 @@
       <UkSpinner @ratio={{1}} />
     </div>
   {{else}}
-    <table class="uk-table uk-table-divider uk-table-striped">
+    <UkButton
+      class="uk-button uk-button-primary uk-float-right uk-margin-small-top"
+      {{on "click" this.exportTable}}
+    >
+      {{t "caluma.analytics.preview.export"}}
+    </UkButton>
+    <table
+      class="uk-table uk-table-divider uk-table-striped"
+      id="reports-table"
+    >
       <thead>
         <tr>
           {{#each this.data.value.headings as |header|}}
--- a/packages/analytics/addon/components/ca-report-preview.js
+++ b/packages/analytics/addon/components/ca-report-preview.js
@@ -1,9 +1,12 @@
+import { action } from "@ember/object";
+import { next } from "@ember/runloop";
 import { inject as service } from "@ember/service";
 import Component from "@glimmer/component";
 import { tracked } from "@glimmer/tracking";
 import { queryManager } from "ember-apollo-client";
 import { task } from "ember-concurrency";
 import { trackedTask } from "reactiveweb/ember-concurrency";
+import * as XLSX from "xlsx";
 
 import getAnalyticsResultsQuery from "@projectcaluma/ember-analytics/gql/queries/get-analytics-results.graphql";
 
@@ -57,5 +60,18 @@ export default class CaReportPreviewComponent extends Component {
       }
     }
     return null;
+  }
+
+  @action
+  exportTable() {
+    next(() => {
+      const wb = XLSX.utils.table_to_book(
+        document.getElementById("reports-table"),
+      );
+      XLSX.writeFile(
+        wb,
+        `${new Date().toLocaleDateString()}_${this.args.slug}.xlsx`,
+      );
+    });
   }
 }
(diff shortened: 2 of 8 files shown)


packages/analytics/addon/components/ca-report-preview.js
```diff
@@ -58,4 +61,14 @@ export default class CaReportPreviewComponent extends Component {
...
     return null;
   }
+
+  @action
+  exportTable() {
+    next(() => {
+      const wb = XLSX.utils.table_to_book(
+        document.getElementById("reports-table"),
+      );
+      XLSX.writeFile(wb, `${new Date().toLocaleDateString()}.xlsx`);
```
Yelinz, line 71: maybe include `this.args.slug` in the name, seems odd to only have a date

packages/analytics/package.json
```diff
@@ -35,7 +35,8 @@
     "ember-uikit": "^9.1.1",
     "ember-validated-form": "^7.0.1",
     "graphql": "^15.8.0",
-    "reactiveweb": "^1.2.2"
+    "reactiveweb": "^1.2.2",
+    "xlsx": "https://cdn.sheetjs.com/xlsx-0.20.2/xlsx-0.20.2.tgz"
```
Yelinz, line 39: Pretty odd package, do you have any gripes with this @anehx ? 
Would it make sense to have this in the backend?

Yelinz (commented):
Could we get a screenshot to see how the button looks like in the UI?

packages/analytics/package.json
open-dynaMIX: Yes, it would make sense to do this in the backend. The problem is, our current analytics approach does not scale. Implementing a new endpoint in core caluma for something we know we'll have to get rid of just doesn't make any sense.
The approach here is pretty much copy/paste from something @StephanH90 did in another project.

packages/analytics/addon/components/ca-report-preview.js
open-dynaMIX: Done

packages/analytics/addon/components/ca-report-preview.hbs
```diff
@@ -4,7 +4,17 @@
...
     </div>
   {{else}}
-    <table class="uk-table uk-table-divider uk-table-striped">
+    <UkButton
+      @label={{t "categorySelection.loadMore"}}
+      class="uk-button uk-button-primary uk-float-right uk-margin-small-top"
+      {{on "click" this.exportTable}}
+    >
+      {{t "caluma.analytics.preview.export"}}
+    </UkButton>
```
Yelinz, line 13: Would probably look better when moved to https://github.com/projectcaluma/ember-caluma/blob/main/packages/analytics/addon/templates/reports/edit.hbs next to the delete button. Moved logic should still work, as the ID is still the same.

Also `@label` needs to be removed as it does nothing, when setting the value in the element.

open-dynaMIX: > Would probably look better when moved to https://github.com/projectcaluma/ember-caluma/blob/main/packages/analytics/addon/templates/reports/edit.hbs next to the delete button. Moved logic should still work, as the ID is still the same.

This package exports an HTML-table to an XLSX. The table only exists on the `RESULT TABLE` tab. So the button should only be visible on the tab,

> Also `@label` needs to be removed as it does nothing, when setting the value in the element.

Done.

Yelinz: yea seems to looks nicer

StephanH90: Make them vertically center aligned at least! My eyes!

StephanH90: Also it looks strange to have a regular button next to a icon button. I would make them both regular buttons instead.

open-dynaMIX: That's why I just made a quick screenshot and didn't commit anything. I dislike this for 3 reasons (2 of them are already mentioned):

1. vertical alignment (can be fixed)
2. regular button next to icon button
3. most importantly: The delete button refers to the analytics table. The export button refers to the data in the preview. In my feeling they just don't belong together

This is not a hill I want to die on. I can fix 1 & 2 if you think it makes sense.

Yelinz: Just looks a bit odd at the top, but we can just let it be.

Yelinz (approved):

PR #2699, opened 2024-04-23
chore(form): add missing sass dev dependencies

Yelinz:
## Description

dependents seem to have problems with importing the form styles (errors with not found), but not analytics or others. The difference I spotted between was form not having sass as any dependency.


Yelinz:
> Pretty sure that having additional `devDependencies` will do anything as host apps don't know about those..

Yea, I know, but just saw they were missing while having scss styles.
Can be considered just a consistency improvement

Yelinz:
Its a flakey modal, cannot reproduce it consistently locally. Will look into it a bit more

PR #2703, opened 2024-04-25
feat(form): add case infos to jexl context

czosel:
This is convenient when you'd like to write a JEXL expression in a task form attached to some work item, that depends on the case's main documents' form.

Related backend MRs: 
- https://github.com/projectcaluma/caluma/pull/2204
- https://github.com/projectcaluma/caluma/pull/2205

### DIFF
--- a/packages/form/addon/gql/fragments/case-form-and-workflow.graphql
+++ b/packages/form/addon/gql/fragments/case-form-and-workflow.graphql
@@ -1,0 +1,28 @@
+fragment CaseFormAndWorkflow on Case {
+  id
+  workflow {
+    id
+    slug
+  }
+  document {
+    id
+    form {
+      id
+      slug
+    }
+  }
+  family {
+    id
+    workflow {
+      id
+      slug
+    }
+    document {
+      id
+      form {
+        id
+        slug
+      }
+    }
+  }
+}
--- a/packages/form/addon/lib/document.js
+++ b/packages/form/addon/lib/document.js
@@ -198,6 +198,7 @@ export default class Document extends Base {
    * @property {Object} jexlContext
    */
   get jexlContext() {
+    const _case = this.raw.workItem?.case ?? this.raw.case;
     return (
       this.parentDocument?.jexlContext ?? {
         // JEXL interprets null in an expression as variable instead of a
@@ -208,6 +209,14 @@ export default class Document extends Base {
           root: {
             form: this.rootForm.slug,
             formMeta: this.rootForm.raw.meta,
+          },
+          case: {
+            form: _case?.document?.form.slug,
+            workflow: _case?.workflow.slug,
+            root: {
+              form: _case?.family.document?.form.slug,
+              workflow: _case?.family.workflow.slug,
+            },
           },
         },
       }
--- a/packages/form/addon/lib/field.js
+++ b/packages/form/addon/lib/field.js
@@ -413,6 +413,7 @@ export default class Field extends Base {
    * - `form`: Legacy property pointing to the root form.
    * - `info.form`: The form this question is attached to.
    * - `info.formMeta`: The meta of the form this question is attached to.
+   * - `info.mainCaseForm`: The main cases' form (can give useful context in task forms).
    * - `info.parent.form`: The parent form if applicable.
    * - `info.parent.formMeta`: The parent form meta if applicable.
    * - `info.root.form`: The new property for the root form.
@@ -423,12 +424,16 @@ export default class Field extends Base {
   get jexlContext() {
     const parent = this.fieldset.field?.fieldset.form;
 
+    const rootDocument = this.document.parentDocument ?? this.document;
+
     return {
       ...this.document.jexlContext,
       info: {
         ...this.document.jexlContext.info,
         form: this.fieldset.form.slug,
         formMeta: this.fieldset.form.raw.meta,
+        mainCaseForm:
+          rootDocument.raw.workItem?.case.family.document?.form.slug,
         parent: parent
           ? {
               form: parent.slug,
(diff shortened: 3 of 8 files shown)


packages/form/addon/lib/field.js
```diff
@@ -423,12 +424,16 @@ export default class Field extends Base {
   get jexlContext() {
     const parent = this.fieldset.field?.fieldset.form;
 
+    const rootDocument = this.document.parentDocument ?? this.document;
```
StephanH90, line 427: But this doesn't work for questions in table, does it? In that case the `parentDocument` will refer to the document where the table is, not the `parentDocument` of the case.

Also, are you sure that the `parentDocument` is set for documents that aren't table rows?

StephanH90 (changes requested):
Just one question.

packages/form/addon/lib/field.js
czosel: As discussed, `parentDocument` is covering questions in table rows, and `document` covers non-table questions.

StephanH90:
Don't forget to rename the property once a decision is made but I will approve now.

StephanH90 (approved):

packages/form/addon/lib/field.js
```diff
@@ -413,6 +413,7 @@ export default class Field extends Base {
    * - `form`: Legacy property pointing to the root form.
    * - `info.form`: The form this question is attached to.
    * - `info.formMeta`: The meta of the form this question is attached to.
+   * - `info.mainCaseForm`: The main cases' form (can give useful context in task forms).
```
anehx, line 416: Might make sense to extend this a little bit:

- `info.case.form`
- `info.case.workflow`
- `info.root.case.form`
- `info.root.case.workflow`

What do you think?

StephanH90: Me likey!

czosel: 👍 good idea!

czosel: After implementing the backend part, I think it should be `case.root` instead of `root.case`, because `root` refers to the currently open "root document" (task form), not to the root case document (case form):
- `info.form` => `document.form.slug`
- `info.root.form` => `document.family.form.slug`
- `info.case` => `document.work_item.case`
- `info.case.root` => `document.work_item.case.family`
- `info.case.root.form` => `document.work_item.case.family.form.slug`

czosel: @anehx added a commit that also makes this feature work with case forms (instead of just task forms) :+1:

StephanH90:
But the naming between the backend and the frontend doesn't match, does it?

In the backend we now have a `main_case_form` which would be the `case.root.form` in the frontend, correct?
(https://github.com/projectcaluma/caluma/pull/2204/files#diff-4200093d628340ea25db028f99d1543cb5c63af275a4a670a9c08adee6444effR150-R158)

Should we try to keep the naming consistent between frontend and backend?

czosel:
> But the naming between the backend and the frontend doesn't match, does it?
> 
> In the backend we now have a `main_case_form` which would be the `case.root.form` in the frontend, correct? (https://github.com/projectcaluma/caluma/pull/2204/files#diff-4200093d628340ea25db028f99d1543cb5c63af275a4a670a9c08adee6444effR150-R158)
> 
> Should we try to keep the naming consistent between frontend and backend?

The backend MR was merged a little bit too quickly, here is the follow-up which should line up with the frontend part: https://github.com/projectcaluma/caluma/pull/2205

packages/form/addon/gql/fragments/case-form-and-workflow.graphql
```diff
@@ -0,0 +1,24 @@
+fragment CaseFormAndWorkflow on Case {
+  id
+  workflow {
+    id
+  }
+  document {
+    id
+    form {
+      id
```
czosel, line 9: we probably need the slug instead

packages/form/addon/lib/document.js
```diff
@@ -198,6 +198,7 @@ export default class Document extends Base {
    * @property {Object} jexlContext
    */
   get jexlContext() {
+    const _case = this.raw.workItem?.case || this.raw.case;
```
czosel, line 201:
```suggestion
    const _case = this.raw.workItem?.case ?? this.raw.case;
```

packages/form/addon/lib/document.js
```diff
@@ -209,6 +210,16 @@ export default class Document extends Base {
...
+          case: _case
+            ? {
+                form: _case.document?.form.id,
+                workflow: _case.workflow.id,
+                root: {
+                  form: _case.family.document?.form.id,
+                  workflow: _case.family.workflow.id,
+                },
+              }
+            : null,
```
czosel, line 222:
```suggestion
          case: {
                form: _case?.document?.form.id,
                workflow: _case?.workflow.id,
                root: {
                  form: _case?.family.document?.form.id,
                  workflow: _case?.family.workflow.id,
                },
              }
```

packages/form/addon/gql/fragments/case-form-and-workflow.graphql
czosel: fixed

packages/form/addon/lib/document.js
czosel: fixed

packages/form/addon/lib/document.js
czosel: The previous implementation also worked for unlinked table rows, but I refactored it as suggested anyway.

packages/form/tests/unit/lib/data.js
```diff
@@ -236,9 +236,48 @@ const answers = {
...
-export default {
+const _case = {
+  id: id("Case"),
+  workflow: {
+    id: "child-case-workflow",
+  },
+  document: {
+    id: id("Document"),
+    form: {
+      id: "child-case-form",
```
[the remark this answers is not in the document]
czosel: fixed

czosel:
@anehx I added another test for unlinked documents. Want to take another look?

anehx (approved):
:100: :rocket:

PR #2708, opened 2024-04-26
fix(number-separator): disallow empty input for mandatory questions

JohnL17:


packages/form/addon/components/cf-field/input/number-separator.js
```diff
@@ -35,12 +35,10 @@ export default class CfFieldInputNumberSeparatorComponent extends Component {
     // We need to remove the thousand separator and replace the decimal
     // separator with a dot in order to parse it into a number. Which character
     // those are is determined per locale in the getters above.
-    const serialized = Number(
```
czosel, line 38: Quickly discussed this with @anehx - right now we're returning `NaN` instead of `null` when a non-numeric string like "abc" is entered - which we shouldn't. 

Would we maybe get the behavior we want by replacing `Number(...)` with `parseFloat(...)`, and leaving everything else unchanged?

JohnL17: I replaced `Number(...)` with `parseFloat(...)` and I get `null` when a non-numeric string is entered.
So is it right then if I enter "abc" the error mesage is `This field can't be blank`?

czosel: As discussed: It's not ideal, but fixing this properly would require some refactoring. Let's go for this for now.

czosel (approved):

PR #2720, opened 2024-05-14
PoC: compare documents

czosel:
@anehx I tried to hack together a quick PoC that shows how me might approach the "historical diff" feature for documents.
image


anehx:
Superseeded by #2999

PR #2723, opened 2024-06-06
fix(distribution): remove suggestions from types and defaultTypes

StephanH90:
## Description

Because we `merge` the default options with any custom supplied options, it is impossible to hide the `suggestions` from the distribution. One of our customers has the situation where a given service should **not** see the `suggestions` because they would always be empty for them.

Because I remove something from the default, this change is potentially breaking.

### DIFF
--- a/packages/distribution/addon/config.js
+++ b/packages/distribution/addon/config.js
@@ -96,14 +96,8 @@ export default function config(target, property) {
           },
           new: {
             defaultDeadlineLeadTime: 30,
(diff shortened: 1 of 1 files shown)


anehx:
Yes there is an option: 

```js
types: {
  suggestions: {
    disabled: true
  }
}
```

StephanH90:
But if we didn't add the suggestions by default by deepmerging the settings, we wouldn't even need this option?

PR #2728, opened 2024-06-21
fix(form): use ember textarea component to correctly rerender values

Yelinz:
## Description

When `field.refreshAnswer` is called a new answer value gets set.
But for HTML textareas the value between the tags `<textarea>foo</textarea>` only acts as a default. To update the actual value one would have to update the `value` attribute through javascript something like `element.value = "bar"`.
This does not happend currently, the rerendering does not happend for `<textarea>` after `refreshAnswer` but for normal `<input>` it works.

Integration test of this always passes, no matter if its plain html textarea or ember textarea. Actually doing it over the UI in an app, causes the new value not be displayed.

Let me know if there is an better fix for this than using the ember component.

### DIFF
--- a/packages/form/addon/components/cf-field/input/textarea.hbs
+++ b/packages/form/addon/components/cf-field/input/textarea.hbs
@@ -8,6 +8,7 @@
   minlength={{@field.question.raw.textareaMinLength}}
   maxlength={{@field.question.raw.textareaMaxLength}}
   readonly={{@disabled}}
+  value={{@field.answer.value}}
   {{on "input" this.input}}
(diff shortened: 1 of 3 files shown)


anehx:
Passing `value={{ @field.answer.value }}` to the textarea is not an option?

Yelinz:
Works, but didnt think that it would. Value is not documented to be working anywhere.

anehx:
Interesting, you are absolutely right it is not documented.. However, since we are using this on multiple places and it works, we should use it ^^

anehx (approved):

PR #2732, opened 2024-07-11
fix(table): improve table disabled

Yelinz:
image

changes text and icon to not be edit when disabled.
stops triggering validation when disabled.


czosel (approved):
:+1:

PR #2744, opened 2024-07-29
fix(distribution): await apollo refetch promises in tests

anehx:
This seems to be necessary since Ember.js v5.10.0


anehx:
This seems to be the same issue as https://github.com/emberjs/ember.js/issues/20724

anehx:
Offline approved by @czosel

anehx:
https://github.com/emberjs/ember.js/pull/20726 should fix this. Let's revert this PR as soon as ember v5.10.2 is released

PR #2745, opened 2024-07-30
Fix various small deprecations

anehx:


anehx:
Offline approved by @czosel

PR #2748, opened 2024-07-30
test(analytics): fix engine testing

anehx:


anehx:
Offline approved by @czosel

PR #2751, opened 2024-07-30
chore(deps): update ember-intl 

anehx:
BREAKING CHANGE: `ember-caluma` now requires the usage of `ember-intl` v7.


anehx:
Offline approved by @czosel

PR #2796, opened 2024-09-06
chore(translations): update italian translations

JohnL17:


packages/distribution/translations/it.yaml
```diff
@@ -68,7 +68,7 @@ caluma:
       search: "Cerca..."
       suggestions: "Raccomandazioni"
       empty: "Non sono stati trovati servizi"
-      groups: "{count} {count, plural, =1 {servizio} other {servizi}}"
+      groups: "{count} {count, plural,= 1 {servizio} other {servizi}}"
```
czosel, line 71: Did you try to start the application? I vaguely remember having had to fix this one during the last import.

JohnL17: You were right. Now that I have changed it, the tests run through. :+1:

Yelinz (approved):

PR #2798, opened 2024-09-12
Add build to test

Yelinz:
Deploy to gh-pages on main fails due to build error, this should catch those errors in the future.


anehx:
Why do we need `corepack` for the husky hooks? Not everyone is using corepack and this might break their setup..

Yelinz:
I think it was because `packageManager` is set in package.json which breaks setups when `corepack` is installed but not used (because pnpm is used directly). It will complain the version is not matching and not run the script.

anehx (approved):

PR #2836, opened 2024-11-05
fix(analytics): preserve whitespace for exports

Yelinz:
## Description

exports do not include white space currently

This should fix that

### DIFF
--- a/packages/analytics/addon/components/ca-report-preview.hbs
+++ b/packages/analytics/addon/components/ca-report-preview.hbs
@@ -30,7 +30,10 @@
         {{#each this.data.value.fields as |row|}}
           <tr>
             {{#each row as |entry|}}
-              <td data-t={{this.getXLSXType entry.value}}>
+              <td
+                data-t={{this.getXLSXType entry.value}}
+                data-v={{this.cleanValue entry.value}}
+              >
                 {{entry.value}}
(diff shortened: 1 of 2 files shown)


packages/analytics/addon/components/ca-report-preview.hbs
```diff
@@ -30,7 +30,10 @@
         {{#each this.data.value.fields as |row|}}
           <tr>
             {{#each row as |entry|}}
-              <td data-t={{this.getXLSXType entry.value}}>
+              <td
+                data-t={{this.getXLSXType entry.value}}
+                data-v={{this.cleanValue entry.value}}
```
derrabauke, line 35: Where are those attributes consumed? Couldn't find any docs. :man_shrugging:

Yelinz: Used in the excel export of the analytics table. Definition is here https://docs.sheetjs.com/docs/csf/cell

derrabauke: Thanks! After skimming over the dox I've found this one to be even more appropriate: https://docs.sheetjs.com/docs/api/utilities/html#implementation-details

derrabauke (approved):

PR #2837, opened 2024-11-08
fix(analytics): display new lines in table

Yelinz:
## Description

Display new lines in analytics table. As it is not openable like a normal caluma table.
Without the change the newline information would be entirely lost on the analytics user.

New look
image

### DIFF
--- a/packages/analytics/addon/components/ca-report-preview.hbs
+++ b/packages/analytics/addon/components/ca-report-preview.hbs
@@ -33,6 +33,7 @@
               <td
                 data-t={{this.getXLSXType entry.value}}
                 data-v={{this.cleanValue entry.value}}
+                class={{this.getCSSClass entry.value}}
               >
                 {{entry.value}}
               </td>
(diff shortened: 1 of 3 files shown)


packages/analytics/addon/components/ca-report-preview.hbs
```diff
@@ -33,6 +33,7 @@
               <td
                 data-t={{this.getXLSXType entry.value}}
                 data-v={{this.cleanValue entry.value}}
+                class={{this.getCSSClass entry.value}}
```
velrest, line 36: Nitpick:
I would prefer if this were just a helper since it doesn't require any context.
```suggestion
                class={{optional-multiline entry.value}}
```

Yelinz: This already is a helper function
https://guides.emberjs.com/v4.12.0/components/helper-functions/#toc_local-helper-functions

I just think calling it with this makes it clear its a local helper

velrest (approved):

PR #2864, opened 2024-12-06
feat(distribution): show tooltip for disabled distrubution types

StephanH90:
## Description

Adds the option to disable buttons in the distribution with a tooltip instead of just hiding the button.

The translations are taken from the `caluma-options` service.

grafik

### DIFF
--- a/packages/distribution/addon/components/cd-inquiry-new-form/select.hbs
+++ b/packages/distribution/addon/components/cd-inquiry-new-form/select.hbs
@@ -1,14 +1,22 @@
 {{#unless this.showAllServices}}
   <div class="uk-margin-bottom uk-button-group" data-test-group-toggle-bar>
     {{#each-in this.config.new.types as |slug config|}}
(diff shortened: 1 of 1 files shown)


packages/distribution/addon/components/cd-inquiry-new-form/select.hbs
```diff
@@ -1,14 +1,22 @@
 {{#unless this.showAllServices}}
   <div class="uk-margin-bottom uk-button-group" data-test-group-toggle-bar>
     {{#each-in this.config.new.types as |slug config|}}
-      {{#unless config.disabled}}
+      {{#if config.notAvailableText}}
```
Yelinz, line 4: The config is implicitly `disabled` by setting `notAbailableText` because you dont check it.
Would make sense to check `disabled` so buttons can be enabled without having to remove the text.

StephanH90:
In the end the customer went with a different version anyway so I'm gonna close this MR.

PR #2871, opened 2025-01-06
Feature/add mandatory asterisk

tstoppe89:
## Description

added mandatory asterisk configuration option for labels of form questions

### DIFF
--- a/packages/form/addon/components/cf-field/label.hbs
+++ b/packages/form/addon/components/cf-field/label.hbs
@@ -3,11 +3,18 @@
     {{@field.question.raw.label}}
   </span>
 
-  {{#if (and (not @field.question.isCalculated) @field.optional)}}
-    <span
-      class="uk-margin-small-left uk-text-muted uk-text-lowercase uk-text-normal"
-    >({{t "caluma.form.optional"}})</span>
-  {{/if}}
+  {{#unless @field.question.isCalculated}}
+    {{#if (and @field.optional (not @useMandatoryAsterisk))}}
+      <span
+        class="uk-margin-small-left uk-text-muted uk-text-lowercase uk-text-normal"
+      >({{t "caluma.form.optional"}})</span>
+    {{else if (and (not @field.optional) @useMandatoryAsterisk)}}
+      <span
+        title={{t "caluma.form.mandatory"}}
+        class="uk-text-muted uk-text-lowercase uk-text-normal mandatory-asterisk"
+      >* </span>
+    {{/if}}
+  {{/unless}}
 
   {{yield}}
 </label>
--- a/packages/form/addon/components/cf-field/label.hbs
+++ b/packages/form/addon/components/cf-field/label.hbs
@@ -11,7 +11,7 @@
     {{else if (and (not @field.optional) this.useMandatoryAsterisk)}}
       <span
         title={{t "caluma.form.mandatory"}}
-        class="uk-text-muted uk-text-lowercase uk-text-normal mandatory-asterisk"
+        class="uk-text-muted uk-text-lowercase uk-text-normal cf-label__mandatory-asterisk"
       >* </span>
     {{/if}}
   {{/unless}}
--- a/packages/form/addon/components/cf-field/label.hbs
+++ b/packages/form/addon/components/cf-field/label.hbs
@@ -4,11 +4,11 @@
   </span>
 
   {{#unless @field.question.isCalculated}}
-    {{#if (and @field.optional (not @useMandatoryAsterisk))}}
+    {{#if (and @field.optional (not this.useMandatoryAsterisk))}}
       <span
         class="uk-margin-small-left uk-text-muted uk-text-lowercase uk-text-normal"
       >({{t "caluma.form.optional"}})</span>
-    {{else if (and (not @field.optional) @useMandatoryAsterisk)}}
+    {{else if (and (not @field.optional) this.useMandatoryAsterisk)}}
       <span
         title={{t "caluma.form.mandatory"}}
         class="uk-text-muted uk-text-lowercase uk-text-normal mandatory-asterisk"
--- a/packages/form/addon/components/cf-field/label.js
+++ b/packages/form/addon/components/cf-field/label.js
@@ -15,8 +15,6 @@ export default class CfFieldLabelComponent extends Component {
     const { USE_MANDATORY_ASTERISK = false } =
       this.config["ember-caluma"] || {};
 
-    return [true, false].includes(this.args?.useMandatoryAsterisk)
-      ? this.args?.useMandatoryAsterisk
-      : USE_MANDATORY_ASTERISK;
+    return this.args?.useMandatoryAsterisk ?? USE_MANDATORY_ASTERISK;
   }
 }
--- a/packages/form/addon/components/cf-field/label.js
+++ b/packages/form/addon/components/cf-field/label.js
@@ -15,6 +15,6 @@ export default class CfFieldLabelComponent extends Component {
     const { USE_MANDATORY_ASTERISK = false } =
       this.config["ember-caluma"] || {};
 
-    return this.args?.useMandatoryAsterisk ?? USE_MANDATORY_ASTERISK;
+    return this.args.useMandatoryAsterisk ?? USE_MANDATORY_ASTERISK;
   }
 }
--- a/packages/form/addon/components/cf-field/label.js
+++ b/packages/form/addon/components/cf-field/label.js
@@ -1,0 +1,22 @@
+import { getOwner } from "@ember/application";
+import Component from "@glimmer/component";
+
+/**
+ * Label component of the CfField
+ *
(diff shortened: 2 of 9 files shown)


czosel (approved):

czosel:
@anehx do you want to take a look at this one as well?

packages/form/addon/components/cf-field/label.hbs
```diff
@@ -3,11 +3,18 @@
...
-  {{/if}}
+  {{#unless @field.question.isCalculated}}
+    {{#if (and @field.optional (not this.useMandatoryAsterisk))}}
+      <span
+        class="uk-margin-small-left uk-text-muted uk-text-lowercase uk-text-normal"
+      >({{t "caluma.form.optional"}})</span>
+    {{else if (and (not @field.optional) this.useMandatoryAsterisk)}}
+      <span
+        title={{t "caluma.form.mandatory"}}
+        class="uk-text-muted uk-text-lowercase uk-text-normal mandatory-asterisk"
```
anehx, line 14: I think the class `mandatory-asterisk` is obsolete, right?

packages/form/addon/components/cf-field/label.hbs
```diff
@@ -3,11 +3,18 @@
...
+      >* </span>
```
anehx, line 15: Is the whitespace afterwards intended?

packages/form/addon/components/cf-field/label.js
```diff
@@ -0,0 +1,22 @@
...
+    return getOwner(this).resolveRegistration("config:environment");
+  }
+
+  get useMandatoryAsterisk() {
+    const { USE_MANDATORY_ASTERISK = false } =
+      this.config["ember-caluma"] || {};
+
+    return [true, false].includes(this.args?.useMandatoryAsterisk)
+      ? this.args?.useMandatoryAsterisk
+      : USE_MANDATORY_ASTERISK;
```
anehx, line 20:
```suggestion
    return this.args.useMandatoryAsterisk ?? USE_MANDATORY_ASTERISK
```

Should almost do the same and would be much easier to read. Or did I miss anything (besides requiring a boolean value to be passed)?

packages/form/tests/integration/components/cf-field/label-test.js
```diff
@@ -34,6 +34,7 @@ module("Integration | Component | cf-field/label", function (hooks) {
       { raw, owner: this.owner },
     );
     this.set("field", document.fields[0]);
+    this.set("useMandatoryAsterisk", false);
```
anehx, line 37: That's the default, do we really need it?

packages/form/tests/integration/components/cf-field/label-test.js
```diff
@@ -48,12 +49,38 @@ module("Integration | Component | cf-field/label", function (hooks) {
   test("it marks optional fields as such", async function (assert) {
     assert.expect(2);
 
-    await render(hbs`<CfField::Label @field={{this.field}} />`);
+    await render(
+      hbs`<CfField::Label @field={{this.field}} @useMandatoryAsterisk={{this.useMandatoryAsterisk}} />`,
+    );
```
anehx, line 54: Let's drop this change as the use case didn't change

packages/form/tests/integration/components/cf-field/label-test.js
```diff
@@ -48,12 +49,38 @@ module("Integration | Component | cf-field/label", function (hooks) {
...
+    assert.expect(1);
+    this.set("useMandatoryAsterisk", true);
+
+    await render(
+      hbs`<CfField::Label @field={{this.field}} @useMandatoryAsterisk={{this.useMandatoryAsterisk}} />`,
+    );
+    this.set("field.question.raw.isRequired", "false");
+
+    assert.dom("label").hasText("Test");
+  });
```
anehx, line 85: This is quite a verbose test for a really simple use-case. I'd do something like this:

```suggestion
  test("it marks mandatory fields as such if useMandatoryAsterisk is configured", async function (assert) {
    assert.expect(1);

    await render(
      hbs`<CfField::Label @field={{this.field}} @useMandatoryAsterisk={{true}} />`,
    );
    
    this.set("field.question.raw.isRequired", "true");
    assert.dom("label").hasText("Test *");
    
    this.set("field.question.raw.isRequired", "false");
    assert.dom("label").hasText("Test");
  });
```

anehx (changes requested):

anehx:
Also what I forgot in my review: We test if it works when that option is passed directly to the component which is very unlikely to happen but not whether the actual configuration of that feature using `USE_MANDATORY_ASTERISK` works.

packages/form/addon/components/cf-field/label.hbs
tstoppe89: Was intended, yes. To make it more comfortable for the user, who's trying to hover it to display the title. If you find it cleaner - I easily can remove the space (should be fine)

packages/form/addon/components/cf-field/label.hbs
tstoppe89: Without any other context, this class is useless, right. I added it, so that a user of caluma can provide specific css styles for that asterisk without wirting crazy css selectors.

It was just an idea to add it like that, do you have any suggestions on it or should I remove it?

packages/form/addon/components/cf-field/label.js
tstoppe89: Would be more beautiful, I agree, but if someone mistakenly set it with "false" as string, it will be true - so i wanted to make sure, that this argument should be a true/false

packages/form/addon/components/cf-field/label.hbs
anehx: That is a really valid point I didn't think about. In that case we should keep it but rename the class to be consistent with other custom classes we use.

In general we *try* to use [BEM](https://getbem.com/) as our naming convention. I'd propose to use `.cf-label__manatory-asterisk` as class name.

packages/form/addon/components/cf-field/label.hbs
anehx: Fair enough, let's keep it.

packages/form/addon/components/cf-field/label.js
```diff
@@ -0,0 +1,20 @@
...
+export default class CfFieldLabelComponent extends Component {
+  get config() {
+    return getOwner(this).resolveRegistration("config:environment");
+  }
+
+  get useMandatoryAsterisk() {
+    const { USE_MANDATORY_ASTERISK = false } =
+      this.config["ember-caluma"] || {};
+
+    return this.args?.useMandatoryAsterisk ?? USE_MANDATORY_ASTERISK;
```
anehx, line 18:
```suggestion
    return this.args.useMandatoryAsterisk ?? USE_MANDATORY_ASTERISK;
```

`this.args` is always an empty object in glimmer components even if not a single argument is passed.

packages/form/tests/integration/components/cf-field/label-test.js
```diff
@@ -56,4 +56,39 @@ module("Integration | Component | cf-field/label", function (hooks) {
...
   });
+
+  test("it marks mandatory fields as such if useMandatoryAsterisk is configured in attribute", async function (assert) {
+    assert.expect(2);
+    this.set("useMandatoryAsterisk", true);
+
+    await render(
+      hbs`<CfField::Label
+  @field={{this.field}}
+  @useMandatoryAsterisk={{this.useMandatoryAsterisk}}
```
anehx, line 67:
```suggestion

    await render(
      hbs`<CfField::Label
  @field={{this.field}}
  @useMandatoryAsterisk={{true}}
```

You can just pass the value directly as `this.useMandatoryAsterisk` doesn't change in this test anyways. I feel like this is much easier to read as you don't have to search for that assignment this way.

anehx (changes requested):
Some **really** minor changes are still needed but in general this looks great!

packages/form/tests/integration/components/cf-field/label-test.js
anehx: Tests look great now!

packages/form/tests/integration/components/cf-field/label-test.js
anehx: I think it's best to not explicitly set default values in tests as it makes the code less readable. To me it's much more important that the non-default value has it's dedicated and explicit test which is the case here :smile:

packages/form/addon/components/cf-field/label.js
anehx: I totally see where this is coming from. However, I'd still drop this check as we don't do this anywhere else for the same use-case. I think the proper way to solve or prevent this problem would be using typescript & glint which we don't do at the moment.

anehx (approved):
Perfect! :100: 

This can be merged (and squashed) as soon as the linter is happy.

PR #2898, opened 2025-03-06
fix(form): remove bypassTimeout from field save

luytena:
Remove logic to bypass timeout during a field save. 
Multiple quick consecutive saves can happen with
every question type, and the debouncing timeout
makes errors due to multiple answer creation
requests less likely.


winged (approved):
LGTM (as far as I can tell)

anehx (approved):

PR #2899, opened 2025-03-25
fix(form-builder): fix hidden and required indicators in form-builder

anehx:
This drops the usage of actual JEXL evaluation in the question list of the form builder in favour of static analysis of the expression (true or false). The problem with the evaluation was, that if an expression only uses context (e.g. `info.form`) it didn't throw and error but evaluated to false. This caused such questions to be marked as always visible when in truth, they'd be conditionally hidden.


czosel (approved):
:+1: makes sense!

PR #3018, opened 2026-01-29
fix(document-validity): use documentValidity to check format validators

anehx:


luytena (approved):
Awesome, looks good to me 💯

PR #3031, opened 2026-04-02
fix: resolve parseFloat truncating numbers with localized separators

JohnL17:


anehx (changes requested):
This **needs** a proper test for the exact use case you're fixing.. Without it it's not clear what problem you're doing with this change.

Also, a proper commit message explaining the broken behaviour would be nice.

packages/form/addon/components/cf-field/input/number-separator.js
```diff
@@ -28,27 +31,47 @@ export default class CfFieldInputNumberSeparatorComponent extends Component {
...
     // separator with a dot in order to parse it into a number. Which character
     // those are is determined per locale in the getters above.
-    const serialized = parseFloat(
-      value
-        .replace(new RegExp(`\\${this.thousandSeparator}`, "g"), "")
-        .replace(new RegExp(`\\${this.decimalSeparator}`), "."),
-    );
 
+    const value = event.target.value.replace(/,/g, ".");
+    this.rawInput = value;
```
anehx, line 62: Why do you need this?

packages/form/addon/components/cf-field/input/number-separator.js
```diff
@@ -1,11 +1,14 @@
...
 import { inject as service } from "@ember/service";
 import Component from "@glimmer/component";
+import { tracked } from "@glimmer/tracking";
 import { cached } from "tracked-toolbox";
 
 export default class CfFieldInputNumberSeparatorComponent extends Component {
   @service intl;
 
+  @tracked rawInput = null;
+
```
anehx, line 11: Only properties that should trigger reactivity should be tracked, is this such a property? I only see one usage and I'm not even sure this is needed.

packages/form/addon/components/cf-field/input/number-separator.js
```diff
@@ -28,27 +31,47 @@ export default class CfFieldInputNumberSeparatorComponent extends Component {
     });
   }
 
+  @cached
+  get numberParts() {
+    // Breaks the dummy number into an array of its exact pieces based on the user's active language
+    return new Intl.NumberFormat(this.intl.locale).formatToParts(11111.1);
```
anehx, line 37:
```suggestion
    return new Intl.NumberFormat(this.intl.primaryLocale).formatToParts(11111.1);
```

packages/form/addon/components/cf-field/input/number-separator.js
```diff
@@ -28,27 +31,47 @@ export default class CfFieldInputNumberSeparatorComponent extends Component {
...
     // We need to remove the thousand separator and replace the decimal
     // separator with a dot in order to parse it into a number. Which character
     // those are is determined per locale in the getters above.
-    const serialized = parseFloat(
-      value
-        .replace(new RegExp(`\\${this.thousandSeparator}`, "g"), "")
-        .replace(new RegExp(`\\${this.decimalSeparator}`), "."),
-    );
 
+    const value = event.target.value.replace(/,/g, ".");
```
anehx, line 61: Can't do that, there are locales that use `,` as thousand separator (e.g. en-US)

JohnL17:
Closed in favor of #3043

PR #3043, opened 2026-04-29
fix(number-separator): don't allow not wanted input 

JohnL17:
When a user adds a comma in an integer or float question everything after the comma and the comma itself is getting stripped from the original input. But the user can't see that because it does only show after a reload. Because of that we don't allow unwanted input in those fields anymore. If the user enters a not allowed character 6 times, a notification shows up.

### DIFF
--- a/packages/form/addon/components/cf-field/input/number-separator.js
+++ b/packages/form/addon/components/cf-field/input/number-separator.js
@@ -5,6 +5,10 @@ import { cached } from "tracked-toolbox";
 
 export default class CfFieldInputNumberSeparatorComponent extends Component {
   @service intl;
+  @service notification;
+
+  blockCount = 0;
+  blockNotified = false;
 
   get disabled() {
     return this.args.disabled || this.args.field?.question.isCalculated;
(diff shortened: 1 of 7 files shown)


JohnL17:
French translation is ordered and will be added as soon as we get it from the customer.

JohnL17:
> French translation is ordered and will be added as soon as we get it from the customer.

All translations are added.

packages/form/addon/components/cf-field/input/number-separator.js
```diff
@@ -38,11 +42,58 @@ export default class CfFieldInputNumberSeparatorComponent extends Component {
     return this.intl.formatNumber(1.1).replace(/\p{Number}/gu, "");
   }
 
+  @cached
+  get questionType() {
+    return this.args.field?.question?.raw?.__typename;
```
anehx:
```suggestion
    return this.args.field.questionType;
```

Please stop slapping optional chaining on everything.

packages/form/addon/components/cf-field/input/number-separator.js
```diff
@@ -38,11 +42,58 @@ export default class CfFieldInputNumberSeparatorComponent extends Component {
...
+  @cached
+  get allowedRegex() {
+    const mil = RegExp.escape(this.thousandSeparator);
+    const dec = RegExp.escape(this.decimalSeparator);
+
+    if (this.questionType === "IntegerQuestion") {
+      return new RegExp(`(\\d|${mil})`);
+    }
+    return new RegExp(`(\\d|${mil}|${dec})`);
+  }
```
anehx, line 56: Needs a comment

packages/form/addon/components/cf-field/input/number-separator.js
```diff
@@ -38,11 +42,58 @@ export default class CfFieldInputNumberSeparatorComponent extends Component {
...
+        ? this.notification.warning(
+            this.intl.t(`caluma.form.validation.blockNotification`, {
+              allowedChars: `0-9 ${this.thousandSeparator}`,
+            }),
+          )
+        : this.notification.warning(
+            this.intl.t(`caluma.form.validation.blockNotification`, {
+              allowedChars: `0-9 ${this.thousandSeparator} ${this.decimalSeparator}`,
+            }),
+          );
```
anehx: This can be simplified. The only thing different is the decimal separator..

```suggestion
      const allowedChars = ["0 - 9", this.thousandSeparator]
      
      if (this.questionType === "FloatQuestion") {
        allowedChars.push(this.decimalSeparator)
      }

      this.notification.warning(
        this.intl.t(`caluma.form.validation.blockNotification`, {
          allowedChars: allowedChars.join(" "),
        }),
      )
```

packages/form/addon/components/cf-field/input/number-separator.js
```diff
@@ -38,11 +42,58 @@ export default class CfFieldInputNumberSeparatorComponent extends Component {
...
+    if (event.key.length > 1 || event.ctrlKey || event.metaKey) {
+      return;
+    }
+
+    if (!this.allowedRegex.test(event.key)) {
+      event.preventDefault();
+      this.blockCount += 1;
+    }
+
+    if (this.blockCount > 5 && !this.blockNotified) {
```
anehx, line 72: Add comments explaining this logic

anehx (changes requested):

packages/form/addon/components/cf-field/input/number-separator.js
JohnL17: Done.

packages/form/addon/components/cf-field/input/number-separator.js
JohnL17: Ah nice. I updated that.

packages/form/addon/components/cf-field/input/number-separator.js
JohnL17: Done.

packages/form/addon/components/cf-field/input/number-separator.js
JohnL17: I apparently didn't found the direct way to `questionType` :sweat_smile:. Changed it.

anehx: Now that I see it, I feel like the getter is quite unnecessary as you can just access it directly..

packages/form/addon/components/cf-field/input/number-separator.js
anehx: I'm missing something about the difference between integer and float in the comment

anehx (changes requested):

packages/form/addon/components/cf-field/input/number-separator.js
JohnL17: Done.

packages/form/addon/components/cf-field/input/number-separator.js
JohnL17: Done.

anehx (approved):
💯 Nice work!

PR #3050, opened 2026-05-19
test(number-separator): assert new thousand separator char for de-CH

anehx:
CLDR 48 (Oct 2025) changed the de-CH group separator from U+2019 (’) to U+0027 ('). Chrome/V8 picks this up via its bundled ICU, so `Intl.NumberFormat` output (which we use via `ember-intl`s `formatNumber`) for de-CH now emits a straight apostrophe where it previously emitted a curly one.

CLDR 48.2 (Mar 2026) applied the same change to fr-CH for consistency.

For more information, check the CLDR changelog:

- https://cldr.unicode.org/downloads/cldr-48#numbers-1
- https://cldr.unicode.org/downloads/cldr-48#482-changes


velrest (approved):
Thanks!

PR #3063, opened 2026-06-26
refactor(form-builder): refactor form components

hwangh95:
## Description

This PR cleans up the data fetching in the general view of the `form-builder` and refactors how the `code-editor` component loads and updates the third-party CodeJar module

Fixes the issue where `form-builder` would crash when switching slugs, since the `code-editor` component didn't remove the observer before unloading

## Dependencies

### DIFF
--- a/packages/form-builder/addon/components/cfb-code-editor.js
+++ b/packages/form-builder/addon/components/cfb-code-editor.js
@@ -1,8 +1,8 @@
 import { action } from "@ember/object";
-import { addObserver } from "@ember/object/observers";
 import Component from "@glimmer/component";
 import { CodeJar } from "codejar";
 import { task, timeout } from "ember-concurrency";
+import { modifier } from "ember-modifier";
 import hljs from "highlight.js/lib/core";
 import json from "highlight.js/lib/languages/json";
 import markdown from "highlight.js/lib/languages/markdown";
@@ -60,14 +60,18 @@ export default class CfbCodeEditorComponent extends Component {
     this.args.update(value);
   }
 
-  updateEditorContent = task({ restartable: true }, async () => {
+  updateEditorContent = modifier((_, [value]) => {
+    this.updateEditorContentTask.perform(value);
+  });
+
+  updateEditorContentTask = task({ restartable: true }, async (value) => {
     // This function is called everytime `this.args.value` changes. In order to
     // not trigger too many updates, we debounce it for 10ms.
     await timeout(10);
 
-    if (!this._didChange(this.args.value)) return;
+    if (!this._didChange(value)) return;
 
-    this._lastValue = simplify(this.args.value);
+    this._lastValue = simplify(value);
     this._editor.updateCode(this.stringValue, false);
   });
 
@@ -78,15 +82,8 @@ export default class CfbCodeEditorComponent extends Component {
       hljs.highlightElement(editor);
     });
 
-    // set initial value
-    this._editor.updateCode(this.stringValue, false);
-    this._lastValue = simplify(this.args.value);
-
     // register update method
     this._editor.onUpdate(this.onUpdate);
-
-    // eslint-disable-next-line ember/no-observers
-    addObserver(this.args, "value", this.updateEditorContent, "perform");
   }
 
   @action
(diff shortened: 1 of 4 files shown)


packages/form-builder/addon/components/cfb-code-editor.js
```diff
@@ -60,37 +60,29 @@ export default class CfbCodeEditorComponent extends Component {
     this.args.update(value);
   }
 
-  updateEditorContent = task({ restartable: true }, async () => {
+  updateEditorContent = modifier(async (_, [value]) => {
     // This function is called everytime `this.args.value` changes. In order to
     // not trigger too many updates, we debounce it for 10ms.
     await timeout(10);
```
anehx, line 70: This doesn't work if we're not in the context of a restartable concurrency task (see https://ember-concurrency.com/docs/examples/autocomplete for an example) so we probably need to keep the task and just trigger that from your update modifier.

packages/form-builder/addon/components/cfb-code-editor.js
```diff
@@ -60,37 +60,29 @@ export default class CfbCodeEditorComponent extends Component {
...
+    if (!this._didChange(value)) return;
 
-    this._lastValue = simplify(this.args.value);
+    this._lastValue = simplify(value);
     this._editor.updateCode(this.stringValue, false);
   });
 
-  @action
-  didInsertNode(element) {
+  codeEditor = modifier((element) => {
```
anehx: Let's rename this to be more clear what the function is actually doing

```suggestion
  initCodeEditor = modifier((element) => {
```

packages/form-builder/addon/components/cfb-code-editor.js
```diff
@@ -60,37 +60,29 @@ export default class CfbCodeEditorComponent extends Component {
...
 
-    // eslint-disable-next-line ember/no-observers
-    addObserver(this.args, "value", this.updateEditorContent, "perform");
-  }
-
-  @action
-  willDestroyNode() {
-    this._editor.destroy();
-  }
+    // register deconstructor on component destroy
```
anehx:
```suggestion
    // Register destructor on component destroy to make sure the CodeJar instance is destroyed as well
```

anehx (changes requested):

packages/form-builder/addon/components/cfb-code-editor.js
```diff
@@ -60,37 +60,34 @@ export default class CfbCodeEditorComponent extends Component {
     this.args.update(value);
   }
 
-  updateEditorContent = task({ restartable: true }, async () => {
+  updateEditorContent = modifier(async (_, [value]) => {
```
anehx:
```suggestion
  updateEditorContent = modifier((_, [value]) => {
```

No need for `async` if you don't `await` anything.

packages/form-builder/addon/components/cfb-form-editor/general.js
```diff
@@ -19,22 +19,18 @@ export default class CfbFormEditorGeneral extends Component {
 
   Validations = FormValidations;
 
-  get data() {
-    return this._data.value?.[0]?.node;
+  get dataIsLoaded() {
+    return !this.data.value || "slug" in this.data.value;
```
anehx: This warrants a comment. We should explain that `watchQuery` returns an empty `TrackedObject` that is truthy instead of something falsey (or god forbid, an error :sweat_smile: )

packages/form-builder/addon/components/cfb-code-editor.js
anehx: Comment looks good, but you kept the typo: ~~deconstructor~~ -> destructor

`ember-modifier` uses [`registerDestructor`](https://api.emberjs.com/ember/6.12/functions/@ember%2Fdestroyable/registerDestructor) in the background.

anehx (changes requested):

packages/form-builder/addon/components/cfb-form-editor/general.js
hwangh95: Also changing to this, since the other way doesn't make quite sense
```suggestion
    return this.data.value && "slug" in this.data.value;
```

anehx (approved):
Perfect! Thanks for fixing this annoying issue and even refactoring some tech-debt! 💯 🚀

PR #3067, opened 2026-07-08
fix(form-builder): prevent duplicate questions on certain reorders

### LINKED ISSUE #2135: Duplication of question when reordering last 2 questions in form builder list
[Showcase of bug](https://user-images.githubusercontent.com/30687616/193222286-e72ae30f-dec7-46b3-9c3d-389f5d11ceca.webm)

There seems to be a bug when reordering the 2nd last question to the last place. The 2nd last question will be duplicated if its dragged to be the last question. The duplication only happens in that specific condition.

The duplicated question does not actually exist, the next reorder request shows a null in the place of the duplicate.
```json
{
	"operationName": "ReorderFormQuestions",
	"query": "mutation ReorderFormQuestions($input: ReorderFormQuestionsInput!, $search: String) {\n  reorderFormQuestions(input: $input) {\n    form {\n      id\n      ...FormInfo\n      questions(filter: [{search: $search}]) {\n        edges {\n          node {\n            id\n            ...QuestionInfo\n            __typename\n          }\n          __typename\n        }\n        __typename\n      }\n      __typename\n    }\n    __typename\n  }\n}\n\nfragment FormInfo on Form {\n  id\n  name\n  slug\n  description\n  isArchived\n  isPublished\n  __typename\n}\n\nfragment QuestionInfo on Question {\n  id\n  slug\n  label\n  isRequired\n  isHidden\n  meta\n  isArchived\n  infoText\n  __typename\n}\n",
	"variables": {
		"input": {
			"form": "3-questions",
			"questions": [
				"frage",
				"zahl",
				null,
				"3rd"
			]
		},
		"search": ""
	}
}
```

hwangh95:
## Description

Because `key` value wasn't set on the `each` helper for question list, Ember was losing track of which DOM elements belonged to the component when the list was reordered with UIkit's `uk-sortable`.

This caused an unstable rerendering of the component and old elements were not properly destroyed, showing up as duplicate questions.

This lets Ember properly keep track of the DOM elements, and reuses them instead of destroying and recreating them.

Fixes #2135

## Dependencies

### DIFF
--- a/packages/form-builder/addon/components/cfb-form-editor/question-list.hbs
+++ b/packages/form-builder/addon/components/cfb-form-editor/question-list.hbs
@@ -71,7 +71,7 @@
     {{#if this.questionTask.isRunning}}
       <li class="uk-text-center"><UkSpinner @ratio={{1.5}} /></li>
     {{else}}
-      {{#each this.questions as |item|}}
+      {{#each this.questions key="node.slug" as |item|}}
         <CfbFormEditor::QuestionList::Item
           data-test-question-list-item={{item.node.slug}}
           @mode={{this.mode}}


anehx (approved):
😍 Thanks so much for fixing this!

PR #3070, opened 2026-07-30
refactor(form-builder): cleanup question-list and options

hwangh95:
## Description

This cleans up some of the implementations of `question-list` and fixes a reordering bug in `options`

## Dependencies

### DIFF
--- a/packages/form-builder/addon/components/cfb-form-editor/question-list.hbs
+++ b/packages/form-builder/addon/components/cfb-form-editor/question-list.hbs
@@ -41,7 +41,7 @@
         placeholder="{{t 'caluma.form-builder.global.search'}}..."
         aria-label="{{t 'caluma.form-builder.global.search'}}"
         value={{this.search}}
-        {{on "input" (perform this.searchTask)}}
+        {{on "input" this.searchTask.perform}}
       />
     </div>
   {{/if}}
@@ -73,8 +73,8 @@
           @mode={{this.mode}}
           @question={{item.node}}
           @onEditQuestion={{@onEditQuestion}}
-          @onRemoveQuestion={{perform this.removeQuestion}}
-          @onAddQuestion={{perform this.addQuestion}}
+          @onRemoveQuestion={{this.removeQuestion.perform}}
+          @onAddQuestion={{this.addQuestion.perform}}
           @onClickForm={{@onClickForm}}
         />
       {{else}}
--- a/packages/form-builder/addon/components/cfb-form-editor/question-list.hbs
+++ b/packages/form-builder/addon/components/cfb-form-editor/question-list.hbs
@@ -1,6 +1,7 @@
 <div
   ...attributes
-  {{did-insert (queue this.setupUIkit (perform this.questionTask))}}
+  {{on "moved" this._handleMoved}}
+  {{did-insert (perform this.questionTask)}}
   {{did-update (perform this.questionTask) @form}}
 >
   <div class="uk-text-center uk-margin">
@@ -80,8 +81,6 @@
           @onRemoveQuestion={{perform this.removeQuestion}}
           @onAddQuestion={{perform this.addQuestion}}
           @onClickForm={{@onClickForm}}
-          @onRegister={{this.registerChild}}
-          @onUnregister={{this.unregisterChild}}
         />
       {{else}}
         <li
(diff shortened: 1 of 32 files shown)


packages/form-builder/addon/components/cfb-form-editor/question-list.js
```diff
@@ -22,31 +21,30 @@ export default class ComponentsCfbFormEditorQuestionList extends Component {
 
   @tracked _search = "";
   @tracked mode = this.args.mode || "reorder";
-  @tracked _children = [];
-  @tracked cursor = null;
-  @tracked hasNextPage = true;
-  @tracked items = [];
+  @tracked showNextPage = true;
```
anehx: What's the reason for renaming this? If I see correctly, it still holds the information on whether it has a next page. The information whether the "load more" is shown still consists of this property in combination with the correct mode - right?

packages/form-builder/addon/components/cfb-form-editor/question-list.js
```diff
@@ -78,7 +76,7 @@ export default class ComponentsCfbFormEditorQuestionList extends Component {
       );
 
       this.cursor = questions.pageInfo.endCursor;
-      this.hasNextPage = questions.pageInfo.hasNextPage;
+      this.showNextPage = this.hasNextPage = questions.pageInfo.hasNextPage;
```
anehx: Same as above - do we really need to rename it? If so, let's drop `this.hasNextPage` duplication

packages/form-builder/addon/components/cfb-form-editor/question-list.js
```diff
@@ -22,31 +21,30 @@ export default class ComponentsCfbFormEditorQuestionList extends Component {
...
   }
 
-  get questionTaskValue() {
+  get questions() {
     return this.questionTask.lastSuccessful?.value ?? [];
   }
 
+  queryQuestions = modifier(() => {
+    this.questionTask.perform();
+  });
```
anehx: We normally (nowadays) solve "loading async data" with resources. You can use `trackedTask` from `reactiveweb/ember-concurrency` to trigger the `questionTask` when needed.

This pattern with an in-class modifier absolutely works as well, but is not common in our codebases so I'd prefer writing it the way we know it.

```suggestion
  questions = trackedTask(this, this.questionTask, () => []);
```

Then instead of accessing `this.questions` directly, you can use `this.questions.[value,isRunning]` etc. - similar to `trackedFunction`.

https://reactive.nullvoxpopuli.com/variables/ember-concurrency.trackedTask.html#example

This way we could also get rid of the rather weird triggering of the search: If you declare it in the task thunk, the task will be re-triggered as soon as the search value changes.

anehx (changes requested):
Looks good in general, thanks for tackling this. I think with a few tweaks (mainly using resources) we could get the current, rather oldschool code up to speed so our devs may understand it a bit better.

Let me know if you have any questions.

packages/form-builder/addon/components/cfb-form-editor/question-list.js
hwangh95: The issue comes from  the fact that `this.hasNextPage` was tracked. The task depends on the property to perform the task, but also updates it in the same task, so when the value was updated, it would trigger the `questionTask` again and again, until there were no more pages.

packages/form-builder/addon/components/cfb-form-editor/question-list.hbs
```diff
@@ -45,7 +41,7 @@
         placeholder="{{t 'caluma.form-builder.global.search'}}..."
         aria-label="{{t 'caluma.form-builder.global.search'}}"
         value={{this.search}}
-        {{on "input" (fn (mut this.search))}}
+        {{on "input" (perform this.searchTask)}}
```
anehx:
```suggestion
        {{on "input" this.searchTask.perform}}
```

I think at some point there was a deprecation warning but I can't find that anymore. Anyhow, the direct syntax is a bit easier on the eyes.

https://github.com/machty/ember-concurrency/releases/tag/4.0.0

> (Not really an enhancement, but) users are encouraged to directly use a task's bound `.perform` method (and curry with `fn` helper if needed) rather than to use the classic `perform` helper, which doesn't really add much value anymore

anehx (changes requested):
One very minor nitpick, then we can merge this 😍 

This is so much better than before.. Thanks! 💯 🚀

packages/form-builder/addon/components/cfb-form-editor/question-list.hbs
hwangh95: I'll do a cleanup commit and replace all instances of `perform`

anehx (approved):
🎉

PR #3071, opened 2026-08-04
Improve JEXL context

anehx:
### fix(jexl): set info.case to null if the document has no case

The JEXL context always contained a `case` object, even for documents
without a case - all of its properties were simply `undefined`. Checking
for the absence of a case therefore required inspecting a nested
property instead of `info.case` itself.

Only build the object if a case is given and pass `null` otherwise,
which mimics the behaviour of the backend.

### feat(jexl): expose case and work item meta in JEXL context

Add `info.case.meta` and `info.case.root.meta`, and introduce
`info.workItem` with `task` and `meta`.

- [Backend PR](https://github.com/projectcaluma/caluma/pull/2628)
- [Docs PR](https://github.com/projectcaluma/docs/pull/12)

### DIFF
--- a/packages/form/addon/gql/fragments/case-form-and-workflow.graphql
+++ b/packages/form/addon/gql/fragments/case-form-and-workflow.graphql
@@ -1,5 +1,6 @@
 fragment CaseFormAndWorkflow on Case {
   id
+  meta
   workflow {
     id
     slug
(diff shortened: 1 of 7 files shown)


nlzet (approved):
LGTM 👍

PR #3072, opened 2026-08-07
feat(form/tables): add row document to table question before editing

winged:
## Description

Earlier, we only attached a row document to the table upon "saving". This has the problem that any JEXL expression (calc, option visibility, etc) that relies on a question *outside* of the table is not visible in the backend, leading to inconsistent behaviour.

Now, a table row is attached properly directly after creating the row doc, so the backend has all the context to validate and evaluate the row's fields as the user edits it.

Of course, if we "cancel" editing, this now requires two operations as well.

## Dependencies

Not a strict dependency, but this goes along with the following backend change:

https://github.com/projectcaluma/caluma/pull/2633

### DIFF
--- a/packages/form/addon/components/cf-field/input/table.js
+++ b/packages/form/addon/components/cf-field/input/table.js
@@ -74,22 +74,68 @@ export default class CfFieldInputTableComponent extends Component {
       owner,
     });
 
+    // Open the modal immediately - attaching the row to the table (below) is
+    // debounced and would otherwise noticeably delay opening the modal. While
+    // this task runs, the row's fields and the cancel button are disabled in
+    // the modal, so no answer can be saved to (and validated by) the backend
+    // and the row cannot be removed again before it is attached to the table.
     this.documentToEditIsNew = true;
     this.documentToEdit = newDocument;
     this.showAddModal = true;
+
+    // Attach the row document to the table answer, so the backend has the
+    // full context to validate and evaluate the row's fields while editing.
+    // `documentToEditIsNew` marks the row as preliminary, so it will be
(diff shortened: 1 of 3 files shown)


packages/form/addon/components/cf-field/input/table.js
```diff
@@ -74,22 +74,58 @@ export default class CfFieldInputTableComponent extends Component {
       owner,
     });
 
+    // Open the modal immediately - attaching the row to the table (below) is
+    // debounced and would otherwise noticeably delay opening the modal.
```
anehx: That makes sense, nice! However, in that case we need to make sure the document only starts getting rendered as soon as the linking is actually done - otherwise we may have wrong context.

It would probably work right now, but this is a footgun of the future, so let's address it now.

packages/form/addon/components/cf-field/input/table.js
```diff
@@ -74,22 +74,58 @@ export default class CfFieldInputTableComponent extends Component {
...
+      // Attaching failed - close the modal, which removes the preliminary
+      // row again
+      await this.close.perform();
     }
+  });
+
+  async deleteRow(document) {
+    /**
+     * Delete row without asking. Remove from the table, then remove row document
+     */
```
anehx: In JS, docstrings go above the method:

```suggestion
  /**
   * Delete row without asking. Remove from the table, then remove row document
   *
   * @async
   * @method deleteRow
   * @param {Document} document The row document to delete
   * @return {Promise<Void>}
   */
  async deleteRow(document) {
```

anehx (changes requested):

anehx (approved):
Nice work! 🎉

PR #3073, opened 2026-08-07
feat(jexl): take the global JEXL context from the backend

anehx:
## Description

`info.case` and `info.workItem` were built in the frontend from the case and work item fetched alongside the document. But the work item the frontend knows about is not the one the context is about: we look it up for action buttons, so for a case document `workItemUuid` resolves the complete workflow form work item *inside* the case, while the backend puts the work item *above* it - the case's parent work item - into the context. Building the context here meant reusing the action button work item for something it isn't.

`documentGlobalJexlContext` now delivers that part of the context ready made, so `Document.jexlContext` only has to add what depends on the document and the field, and the action button work item stays out of the context entirely. Documents that are not fetched through the document answers query - e.g. the ones constructed for bulk editing - have no global context and fall back to an empty case and work item.

The case and work item are still queried, but only with the fields `workItemUuid` needs. The `CaseFormAndWorkflow` fragment has no users left and is dropped.

## Dependencies

Depends on https://github.com/projectcaluma/caluma/pull/2632


nlzet (approved):
LGTM 👍 Testing it out and works like a charm

PR #3074, opened 2026-08-07
feat(distribution): build the global JEXL context for new inquiries

anehx:
## Description

Documents are evaluated against the global JEXL context the backend delivers, but the bulk edit form has no document to ask one for - it constructs a throwaway document to render the inquiry form of an inquiry that doesn't exist yet. Expressions on that form would see neither the case nor the work item.

So we assemble the context ourselves, for this one case. The case comes from the controls query, extended with the fields it needs. The work item is the one `createInquiry` is about to create: its task is configured, and its meta - the one thing we can't know here - comes from `distributionInquiryWorkItemMeta` on the caluma options service, which
the host app can implement if it creates inquiries with a meta. What comes out has to match what the backend will return once the inquiry exists, otherwise expressions change under the user on save.

## Dependencies

Depends on #3073

### DIFF
--- a/packages/core/addon/services/caluma-options.js
+++ b/packages/core/addon/services/caluma-options.js
@@ -116,6 +116,28 @@ export default class CalumaOptionsService extends Service {
     return DateTime.now().plus({ days: defaultLeadTime }).toISODate();
   }
 
+  /**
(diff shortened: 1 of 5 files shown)


nlzet (approved):
LGTM 👍

PR #3079, opened 2026-08-10
fix(form/tables): don't validate row document when adding

anehx:
The `onSave` UI action will immediately validate the respective row document. Since we just added it and it's empty, this will fail, as soon as the row form has a required question (which it always should). In order to prevent that, we use the private save task of the field itself directly.


winged (approved):
LGTM I think - I did experiment with the fix, but definitely wasn't as clean as your version :-D

PR #3082, opened 2026-08-11
fix(table): don't show form while the row doc is being attached

winged:
We shouldn't abuse the `@disabled` attribute to pass along "loading" state. Some components don't deal well with quick state changes and may behave incorrectly.

Instead, we now show a spinner until the row document is fully attached.

### DIFF
--- a/packages/form/addon/components/cf-field/input/table.hbs
+++ b/packages/form/addon/components/cf-field/input/table.hbs
@@ -124,13 +124,19 @@
     as |modal|
   >
     <modal.body>
-      <CfFormWrapper
-        @document={{this.documentToEdit}}
-        @fieldset={{object-at 0 this.documentToEdit.fieldsets}}
-        @disabled={{or @disabled this.add.isRunning}}
-        @context={{@context}}
-        @compare={{@compare}}
-      />
+      {{#if this.add.isRunning}}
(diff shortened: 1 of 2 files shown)


packages/form/addon/components/cf-field/input/table.hbs
```diff
@@ -124,13 +124,19 @@
...
     <modal.body>
-      <CfFormWrapper
-        @document={{this.documentToEdit}}
-        @fieldset={{object-at 0 this.documentToEdit.fieldsets}}
-        @disabled={{or @disabled this.add.isRunning}}
-        @context={{@context}}
-        @compare={{@compare}}
-      />
+      {{#if this.add.isRunning}}
+        <div class="uk-align-center">
```
anehx:
```suggestion
        <div class="uk-text-center">
```

We don't use the align classes as they still use float.

anehx (changes requested):
Minor nitpick about class selection and the tests are still failing.

packages/form/addon/components/cf-field/input/table.hbs
winged: Fixed, also updated the test to reflect the new behaviour

anehx (approved):
Awesome 🔥 Great work!

PR #3083, opened 2026-08-12
feat(distribution): allow custom icon component for inquiry status

anehx:
Status configs can now define an `iconComponent` which is rendered instead of the default `UIkit` icon. The icon rendering of the inquiry divider and the navigation status indicator is extracted into the new `CdInquiryStatusIcon` component, which either renders the configured component or falls back to the previous `UkIcon` with tooltip.

Color class and tooltip are only applied to the default icon - a custom component receives `@status` and `@tooltipPos`, and is responsible for rendering color and tooltip itself.


StephanH90 (approved):
AY LGTM.

PR #3084, opened 2026-08-12
fix(distribution): apply color and tooltip to custom status icons

anehx:
The color class and the status tooltip are now applied to a configured `iconComponent` as well, not only to the default `UkIcon`. Custom icon components therefore need to spread `...attributes` onto their root element, if they want the default color class and tooltop handling.


StephanH90 (approved):

StephanH90:
2nd time is the charm!

PR #3095, opened 2026-09-18
fix(css): render selected checkboxes and radios in print mode

nlzet:
## Render selected checkboxes and radios in print mode

When printing a page currently the selected checkboxes and radios are not visible, this adds the media print css to fix that.

**Before:**

image

**After:**

image


nlzet:
Moving to ember-uikit instead

